PR #21, opened 2012-04-19
Hello, some feature enhancements and bug fixes to your excellent libhtp

nickfajones:


calfeld-zz:
Thanks, Nick!

It looks good.  There's one minor issue with some API documentation that
seems to be incomplete (see
https://github.com/nickfajones/libhtp/commit/3f57dfec754becc0475d52f702d978eed6d20cdd).
 Once that's addressed, I'll pull it all in.

If it's easier, you can just reply to this e-mail with the doc fix.

c.

On Thu, Apr 19, 2012 at 9:35 AM, Nick Jones <
reply@reply.github.com

> wrote:
> 
> You can merge this Pull Request by running:
> 
>  git pull https://github.com/nickfajones/libhtp master
> 
> Or you can view, comment on it, or merge it online at:
> 
>  https://github.com/ironbee/libhtp/pull/21
> 
> -- Commit Summary --
> - Add response_header_raw members to htp_tx with associated functions
> - Add new hook: hook_response_start
> - Record the response content-type
> - Fix memory leak when response_headers_sep is reallocated after a trailer
> - Changed logic for choosing outgoing address:port
> 
> -- File Changes --
> 
> M htp/htp.h (41)
> M htp/htp_config.c (20)
> M htp/htp_connection_parser.c (10)
> M htp/htp_request.c (24)
> M htp/htp_response.c (38)
> M htp/htp_transaction.c (4)
> M htp/htp_util.c (66)
> 
> -- Patch Links --
> 
>  https://github.com/ironbee/libhtp/pull/21.patch
>  https://github.com/ironbee/libhtp/pull/21.diff
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/ironbee/libhtp/pull/21

## 

Christopher Alfeld | calfeld@qualys.com
Director of Research
Qualys, Inc. | On Demand Security | www.qualys.com
 http://www.qualys.comDirect: 650.801.7776 | Mobile: 608.628.4512

nickfajones:
I've revised the use_local_port commit, but I feel it deserves a bit more discussion.
If you are happy with the first four changes, please pull them and we can discuss the use_local_port feature in isolation.

As it is now, use_local_port it is more a demonstration of the problem, in that libhtp only uses the local connection port as the outgoing connection port, which is not correct when a proxy is deployed as an inline, browser configured proxy, or as a transparent (TPROXY) proxy.

The use_local_port change works well to solve this one issue, but it changes the api of htp_connp_open in a way that I have chosen without consensus.

My feeling is that an author should be able to indicate the personality of a connection when creating the htp_config or the htp_connp instance.  The connection personality is much like the server personality, where the value specified will cause some slight variation of behaviour in libhtp during request processing, such as the way outgoing connection port is chosen.

There may be more 'behaviour' aspects that need to be modelled, thus renaming use_local_port to deployment_personality and treating it as an enumeration rather than just a boolean would be better.  htp.h can have these enum declarations added in the same fashion as the server personalities.

Lets discuss, then I can revise the patch to follow the conclusions.

calfeld-zz:
Hi Nick,

  Ivan Ristic should be in touch with you soon (if not already) regarding
some licensing related issues that came up (for reasons unrelated to you;
but the issues may affect you).

  Brian Rectanus should be in touch with you regarding your e-mail below.

  Thanks,
    c.

On Thu, Apr 19, 2012 at 11:43 PM, Nick Jones <
reply@reply.github.com

> wrote:
> 
> I've revised the use_local_port commit, but I feel it deserves a bit more
> discussion.
> If you are happy with the first four changes, please pull them and we can
> discuss the use_local_port feature in isolation.
> 
> As it is now, use_local_port it is more a demonstration of the problem, in
> that libhtp only uses the local connection port as the outgoing connection
> port, which is not correct when a proxy is deployed as an inline, browser
> configured proxy, or as a transparent (TPROXY) proxy.
> 
> The use_local_port change works well to solve this one issue, but it
> changes the api of htp_connp_open in a way that I have chosen without
> consensus.
> 
> My feeling is that an author should be able to indicate the personality of
> a connection when creating the htp_config or the htp_connp instance.  The
> connection personality is much like the server personality, where the value
> specified will cause some slight variation of behaviour in libhtp during
> request processing, such as the way outgoing connection port is chosen.
> 
> There may be more 'behaviour' aspects that need to be modelled, thus
> renaming use_local_port to deployment_personality and treating it as an
> enumeration rather than just a boolean would be better.  htp.h can have
> these enum declarations added in the same fashion as the server
> personalities.
> 
> Lets discuss, then I can revise the patch to follow the conclusions.
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/ironbee/libhtp/pull/21#issuecomment-5237812

## 

Christopher Alfeld | calfeld@qualys.com
Director of Research
Qualys, Inc. | On Demand Security | www.qualys.com
 http://www.qualys.comDirect: 650.801.7776 | Mobile: 608.628.4512

nickfajones:
Some additional feature and bugfixes, separated into individual commits.  They have been rebased on top of your current HEAD revision.

ivanr:
Thank you, Nick. I am going to review the commits individually, and merge them one by one. If I have any questions, I will most likely ask them here. 0054d5f is already in.

ivanr:
Merged 48be201.

ivanr:
Nick, regarding 857373f, if you want to access chunk data, why wouldn't you use the BODY_DATA callback?

nickfajones:
Hi Ivan,

I can see how 857373f might seem a bit strange as it is based on a need in my situation.

When we receive a BODY_DATA callback, it is giving us a pointer to a block of data and a length, this is true.
But in fact I already have a copy of the body data, in the buffer that I have just fed into libhtp.

As the various callbacks come, I can seek the original buffer by the amount of data that libhtp has consumed: request/response line, then headers.

I wanted the chunked_header_offset so that in the body callbacks I could seek past the chunk header lines and arrive at the pure body data, which can then be isolated into another buffer for scanning, unless the content is compressed, in which case using the htp_tx_data structure is essential.

This chunked header offset mimics the in/out_stream_offset that you already have, which I found very useful for seeking past header and body data.

Please pause on this patch, considering it as an idea for a feature.

In addition to that, on inspection, I realise there could be an issue, as the chunked_header_offset isn't taking into account the '\r\n' before the chunked header, which of course doesn't belong to the pure body data.

ivanr:
Merged  e446eb0 and 784750e.

ivanr:
Merged 0b1ddbc and fac87ed.

ivanr:
Merged the remaining commits (except 857373f). Regarding ed62754, I left it as is for now, but I don't like that we've changed the API. Further, as you suggest, this might be better implemented as a configuration setting. I've made a note to review the behavior related to ports.

PR #22, opened 2012-08-13
Support HOOK_STOP and a NULL ptr deref bugfix.

wxsBSD:
New feature: Add support for HOOK_STOP. When received by a hook the stream
state is set to STOP and the parser will refuse to parse anything more.

Bugfixes:
- Fix comments.
- Check if timestamp is not NULL before doing the memcpy. In my soon to be
  released python bindings we aren't using the timestamp so this was causing
  a crash.


ivanr:
Hi Wesley, and thank you for your submissions! As it is, it's difficult for me to merge your pull request because it combines several unrelated changes. Can you in the future please use one pull request per a number of related changes?

I can offer some feedback, though:
- The timestamp field is not used anyway (the idea was to preserve the timing information, should it be needed), and I am OK with making it optional. However, there are 4 methods that accept a timestamp, and your diff touches only two.
- Regarding HOOK_STOP: when do you need to use it? What is your use case?

wxsBSD:
Sorry about that, I'll make the changes and pull requests discrete in the future.

My diff only touches two places which accept a timestamp because those are the only two places I ran into the problem. I'll fix the other two shortly.

The use case for HOOK_STOP is for a project I'm working on. We have libnids reassembling streams and handing the data to libhtp for HTTP parsing. As hooks are called we have logic in them to decide if we should continue reassembling the stream or not. If we choose not to reassemble the stream we need to signal that to libnids somehow and since it isn't an error condition I thought returning HOOK_ERROR was not the right thing to do. Instead I opted to return HOOK_STOP which is treated differently from an error condition.

ivanr:
Wesley, your new commit touches 2 out of 4 functions, with the remaining 2 functions updated in the old commit (which I cannot merge because it touches a few things). However, because the change is simple, I decided to commit a re-implementation.

wxsBSD:
Thank you. I will rebase my fork off of your latest changes and implement only the HOOK_STOP feature. Once that is done I will issue a new pull request.

wxsBSD:
I issued a new pull request with only the HOOK_STOP feature in a single commit. I guess this one can be closed as the other things it addressed have been handled.

PR #29, opened 2012-09-28
Fix memory leak on cookies with duplicate names

nickfajones:
If a cookie fails to be added to the request_cookies table because
of insert failure or more likely a cookie with the same tag already
exists, then we are currently leaking the memory allocated to store
the the current cookies name and value strings.  This patch fixes
this by evaluating the return value of table_addn,

This fix could also start a discussion about what to do in the case of 
cookies with duplicate tags, present in the request either by incorrect 
header generation by the browser, or the addition of cookies by a 
proxy that unfortunately conflict with the browsers cookies.

In the end, the raw headers are what get written onwards to the server 
by the proxy, if the author uses htp_tx_generate_request_headers_raw,
so the duplicate cookies will be passed on to the server unchanged.
But during processing by libhtp, the duplicate cookie will be skipped, with 
or without this patch, and thus will evade analysis.

It may be wise to at least add a flag to the request transaction with a 
meaning similar to HTP_REQUEST_SMUGGLING, which can 
provide a library user another factor for making a security policy 
decision.

### DIFF
--- a/htp/htp_cookies.c
+++ b/htp/htp_cookies.c
@@ -73,7 +73,10 @@ int htp_parse_single_cookie_v0(htp_connp_t *connp, char *data, size_t len) {
     // Add cookie
     if (connp->cfg->parameter_processor == NULL) {
         // Add cookie directly
-        table_addn(connp->in_tx->request_cookies, name, value);
+        if (table_addn(connp->in_tx->request_cookies, name, value) == -1) {
+            bstr_free(&value);
+            bstr_free(&name);
+        }
     } else {
         // Add cookie through parameter processor
         connp->cfg->parameter_processor(connp->in_tx->request_cookies, name, value);


ivanr:
Are you sure that adding a cookie can fail? The function table_addn will append to the table and will not overwrite any previous value.

nickfajones:
Indeed.  I misinterpreted the code in haste.

So it seems the main subject of my patch is a red-herring.  Although it will protect against memory leaks given an allocation failure inside list_array_push, it has nothing to do with duplicate cookies given the default behaviour of the dslib table implementation.

The default behaviour of dslib table is: duplicate cookie names and values will still be claimed by the request_cookies table, and thus cleaned up when it is cleaned up.   Even though only the first value will be retrieved.

The reason I was seeing this was because I am using the pluggable containers feature to fall back to C++STL container members of the class controlling libhtp.  std::map will not support duplicates, so I retuned -1 from the custom table_addn function to indicate failure to insert.

I think there are two separate issues here, what to do on insertion failure, and what to do with duplicate cookies, I've started to mix them up.

Let me resubmit this patch as a pure safety measure, as it is valid for that purpose, and then create an Enhancement to discuss what to do with duplicate cookies.

ironbee:
On 28 Sep 2012, at 11:19, Nick Jones wrote:

> Let me resubmit this patch as a pure safety measure, as it is valid for that purpose, and then create an Enhancement to discuss what to do with duplicate cookies.

Why are duplicate cookies a problem?

They're perfectly legal, and if they give rise to any security concerns
then those are fully shared by other HTTP request headers, any of which
may be duplicated!

## 

Nick Kew

ivanr:
> The default behaviour of dslib table is: duplicate cookie names and values will still be claimed by the request_cookies > table, and thus cleaned up when it is cleaned up. Even though only the first value will be retrieved.

If you use table_get then you'll see only one Cookie. However, you can (and, depending on the use case, should) iterate over the table to retrieve all Cookies that match the desired name.

PR #30, opened 2012-10-05
Fix htp_tx_generate_request_headers_raw null dereference.

inliniac:
Fix null dereference found in a Suricata unittest DetectHttpRawHeaderTest07 (https://github.com/inliniac/suricata/blob/master/src/detect-http-raw-header.c#L480). The htp_tx_generate_request_headers_raw function is called on an incomplete request from the Suricata detection engine, leading to a null dereference as htp_tx_generate_request_headers_raw assumes tx->request_headers_sep to be initiallized.

### DIFF
--- a/htp/htp_util.c
+++ b/htp/htp_util.c
@@ -2099,7 +2099,8 @@ bstr *htp_tx_generate_request_headers_raw(htp_tx_t *tx) {
         len += bstr_len(hl->line);
     }
 
-    len += bstr_len(tx->request_headers_sep);
+    if (tx->request_headers_sep)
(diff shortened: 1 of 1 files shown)


ivanr:
Victor, did you intend to submit this against 0.3.x? I understand that you are currently using your own branch. In that light, I would fix the master, and then transition Suricata to it when the next major version (of libhtp) is ready.

inliniac:
Yes, I was looking at supporting 0.3.0 just to see what would take, triggered by https://redmine.openinfosecfoundation.org/issues/538

PR #45, opened 2013-07-18
Fix segv caused when we have an invalid host string in host header or

poona:
CONNECT authority.

One way to fix this was manually check if *hostname is NULL before calling htp_validate_hostname(), or inside htp_validate_hostname().  But if we have already set the HTP_HOSTH_INVALID flag, there's no point again trying to validate the hostname and then set the flag again.


poona:
Wrong branch.  Closing PR.

PR #46, opened 2013-07-18
Fix segv caused when we have an invalid host string in host header or

poona:
CONNECT authority.

One way to fix this was manually check if *hostname is NULL before calling htp_validate_hostname(), or inside htp_validate_hostname(). But if we have already set the HTP_HOSTH_INVALID flag, there's no point again trying to validate the hostname and then set the flag again.


inliniac:
Looks like returning HTP_OK causes trouble elsewhere. In htp_transaction.c:490 the hostname ptr is used unconditionally, leading to another segv:
            if ((bstr_cmp_nocase(hostname, tx->request_hostname) != 0) || (port != tx->request_port_number)) {

poona:
Closing and re-updating.

PR #68, opened 2014-02-06
In htp_log, set the htp_log_t::tx field

inliniac:
The htp_log_t structure has a pointer to the transaction in by which
the message was raised. However, the field is never set.

This patch sets it based on either the htp_connp_t::in_tx or the
htp_connp_t::out_tx pointers. The former has preference over the
latter.

### DIFF
--- a/htp/htp_util.c
+++ b/htp/htp_util.c
@@ -370,6 +370,7 @@ void htp_log(htp_connp_t *connp, const char *file, int line, enum htp_log_level_
     log->level = level;
     log->code = code;
     log->msg = strdup(buf);
+    log->tx = connp->in_tx ? connp->in_tx : connp->out_tx;
 
     htp_list_add(connp->conn->messages, log);
 


ivanr:
Your patch is going to associate an incorrect transaction if there is more than one concurrent transaction on a connection. I'd rather leave this as is in 0.5.x, and change the API in 0.6.x so that htp_log() explicitly names the associated transaction.

inliniac:
On 02/10/2014 02:04 PM, Ivan Ristić wrote:

> Your patch is going to associate an incorrect transaction if there is
> more than one transaction on a connection. I'd rather leave this as is
> in 0.5.x, and change the API in 0.6.x so that htp_log() explicitly names
> the associated transaction.

Hmm, but if I don't have this info, I will have the same problem as in
Suricata I will still pick a tx to assign the message to. Is there
really no way of finding out what the active tx is when htp_log is
called? I assumed htp_conn_t would have this info.

ivanr:
When you say "in Suricata", do you mean in a callback? You will always be called with tx supplied. As far as I can tell, all callbacks provide either htp_tx_t or htp_tx_data_t (which has a tx).

inliniac:
No, we first run htp_connp_req_data(), then check the htp_conn_t->messages list to see if any messages were added. So it may be possible that htp_connp_req_data() processed multiple TX' during that period.

I'm not sure how checking this in callbacks would work. We don't use all possible hooks. We need to be able to catch the logs/events from the entire htp processing pipeline.

ivanr:
Then, from your perspective, all the log messages will have the correct tx associated with them. (With the proposed change in 0.6.x.) That's possible, because whoever is logging the message will always know the correct tx. You won't need to worry about it.

inliniac:
If I understand correctly what you are saying, in 0.6.x this won't be an issue, but in 0.5.x it is. Correct?

Any possible work arounds?

ivanr:
Yes, correct. As for the workarounds, I don't think so, not without breaking the API.

PR #70, opened 2014-03-12
test/Makefile.am: Fix issue with --disable-static.

calfeld-zz:
Simple fix to `test/Makefile.am` which allows `make check` to succeed if `--disable-static` is passed to `configure`.  `--disable-static` is nice as it gives (on my machine) ~40% faster compile times by not double-compiling every object file.


ivanr:
Applied to the 0.5.x and 0.6.x branches.

ivanr:
Thanks!

PR #71, opened 2014-04-02
Fix missing files in distribution target

chifflier:
When running autoreconf -f -i on a distributed tarball, the configuration fails with the following error:
    autoreconf -f -i
sh: 1: ./get-version.sh: not found
configure.ac:7: error: AC_INIT should be called with package and version arguments
/usr/share/aclocal-1.14/init.m4:29: AM_INIT_AUTOMAKE is expanded from...
configure.ac:7: the top level

Files VERSION and get-version.sh are required to run autoreconf, on the
sources from a 'make dist' tarball.
Add them to the EXTRA_DIST target.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -4,7 +4,7 @@ ACLOCAL_AMFLAGS = -I m4
 SUBDIRS = htp test docs
 
 DIST_SUBDIRS = htp test docs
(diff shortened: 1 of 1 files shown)


ivanr:
Committed to 0.5.x. (Will be merged into master when 0.5.11 is released.) Thanks!

PR #73, opened 2014-05-15
Fix double Content-Length issue.

wxsBSD:
I've seen situtations where a server will send duplicate Content-Length
headers. I've tested this with Chrome and it will ignore the second header
if the value is the same. If the value is different Chrome throws an error.

This causes a discrepency between how a browser will handle the content
and how libhtp does. To address this I've added support for checking for
duplicate Content-Length headers and ignoring subsequent ones if the value
is the same as the first one.

Also, fix up the test (74-response-multiple-cl) now that this is not an
error condition. Also, add a test where there is a value mismatch which
is a failure.

### DIFF
--- a/htp/htp_response_generic.c
+++ b/htp/htp_response_generic.c
@@ -252,29 +252,51 @@ htp_status_t htp_process_response_header_generic(htp_connp_t *connp, unsigned ch
     // Do we already have a header with the same name?
     htp_header_t *h_existing = htp_table_get(connp->out_tx->response_headers, h->name);
     if (h_existing != NULL) {
-        // TODO Do we want to have a list of the headers that are
-        //      allowed to be combined in this way?
+        /*
+         * While having multiple content-length headers is against the RFC
+         * many browsers will ignore the subsequent ones if the value is
+         * the same.
+         */
+        if (bstr_cmp_c_nocase(h->name, "Content-Length") == 0) {
+            /*
+             * Don't use bstr_cmp() and friends because we want to catch:
+             * Content-Length:12
+             * Content-Length: 12
+             * So parse the content length and compare the integers.
+             */
+            int64_t existing_cl, new_cl;
+            existing_cl = htp_parse_content_length(h_existing->value);
+            new_cl = htp_parse_content_length(h->value);
+            if ((existing_cl == -1 || new_cl == -1) || existing_cl != new_cl) {
+                return HTP_ERROR;
+            } else {
+                h_existing->flags |= HTP_FIELD_REPEATED;
+            }
+        } else {
+            // TODO Do we want to have a list of the headers that are
+            //      allowed to be combined in this way?
 
-        // Add to the existing header.
-        bstr *new_value = bstr_expand(h_existing->value, bstr_len(h_existing->value) + 2 + bstr_len(h->value));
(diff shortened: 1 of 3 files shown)


htp/htp_response_generic.c
```diff
@@ -252,29 +252,51 @@ htp_status_t htp_process_response_header_generic(htp_connp_t *connp, unsigned ch
...
+            /*
+             * Don't use bstr_cmp() and friends because we want to catch:
+             * Content-Length:12
+             * Content-Length: 12
+             * So parse the content length and compare the integers.
+             */
+            int64_t existing_cl, new_cl;
+            existing_cl = htp_parse_content_length(h_existing->value);
+            new_cl = htp_parse_content_length(h->value);
+            if ((existing_cl == -1 || new_cl == -1) || existing_cl != new_cl) {
```
inliniac, line 270: Wonder if this should be an error. Might be better so set an event (flag)?

inliniac:
In the existing code in this case it seems we:
1. expand the stored CL value in "value1,value2"
2. htp_parse_content_length (actually htp_parse_positive_integer_whitespace) will return -1002 on this, as the comma format isn't expected
   -> confirmed this in a test, error message "Invalid C-L field in response: -1002" is set
3. htp_connp_res_data returns HTP_STREAM_ERROR

htp/htp_response_generic.c
wxsBSD, line 270: It should be an error because browsers will not parse the body if the two values do not match. I do not know how a browser would be if there are multiple requests in a session and the first one has mismatched content lengths.

wxsBSD:
Yes, this is exactly what I was seeing.

ivanr:
I think that it's OK to relax response header parsing to allow multiple C-L headers provided the values are the same. But I'd like to spend some time understanding if such a change would open up an undetectable attack vector (I think not). A parsing failure might get some scrutiny (Victor, what do you think?), but a flag is likely to be ignored.

inliniac:
Suricata would not ignore the flag, but in general I understand the concern.

wxsBSD:
I'm not sure how you can set a flag here and continue parsing. If the length of the body is ambiguous how can you parse it and then continue on to the next request? Admittedly, I'm no libhtp expert so maybe I'm missing something. I think the only option is to fail parsing.

inliniac:
In other parts of IDS we use 'personalities' to account for differences like this, in IP defrag for example. If some browsers allow for multiple CL headers, we ideally would support them. Worst case would be that some browsers select the first CL, other the 2nd, etc. I guess more research would be needed before we can say for sure.

Generating an error (which in suri would lead to an alert, if the rule for that case is enabled) should indeed draw an analysts attention.

wxsBSD:
I'll test with Chrome, Safari and Firefox (latest of each) on OS X and report back what they do.

wxsBSD:
In my testing Chrome, Safari and Firefox will stop processing a TCP stream if they are confused about the content length in a transaction. If the content lengths specified are matches none of the browsers I tested care.

ivanr:
Thank you for following up. I will make the necessary changes next week.

ivanr:
Committed to the 0.5.x and 0.6.x branches. There were some memory leaks, which I fixed. Thanks!

PR #75, opened 2014-07-09
Initial travis-ci integration

inliniac:
Enable basic travis-ci integration. This will allow you to enable basic build tests on both commits and pull requests. To complete this, setup your account here: http://docs.travis-ci.com/user/getting-started/

Example from how we use it in Suricata: https://travis-ci.org/inliniac/suricata

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: c
+compiler:
+  - gcc
+  - clang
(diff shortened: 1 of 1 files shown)


inliniac:
If you're going to consider this, you may want to enable the travis-ci account first so that the yml in this PR is tested to work :)

PR #80, opened 2014-10-21
[rfc] Request line leading space handling

inliniac:
Fix handling of leading whitespace on the request line.
Also, make it a per-personality setting.

I wonder if adding the new option to htp_decoder_cfg_t breaks binary compatibility as that is in the htp_cfg_t as an array. Might need to add the option directly to the end of the htp_cfg_t in that case.

Comments are welcome.


inliniac:
After some discussion with @wxsBSD online, we've determined that this breaks the ABI. Will do a new PR.

PR #81, opened 2014-10-22
Request line leading space v1

inliniac:
Same as https://github.com/ironbee/libhtp/pull/80, except now the config part is appended to htp_cfg_t, which shouldn't break the ABI.

cc @wxsBSD


glongo:
Merged after rebase, thanks!

PR #82, opened 2014-10-29
htp_connp_close: handle streams in error state

inliniac:
A stream in the error state would not be handled correctly by
htp_connp_close. The error state would be overwritten to 'CLOSED'
and the final handling of the stream would follow.

This lead to a case where a null pointer dereference could happen.

If for whatever reason inflateInit2 failed, the stream would get
into an error state. By resetting the state, the parser would
later unconditionally invoke the decompressor. As it's setup
failed, this resulted in a null dereference.

This patch makes sure that htp_connp_close doesn't overwrite
the state. The functions called by htp_connp_close are able
to deal with an error state properly.


inliniac:
Thanks Ivan!

PR #86, opened 2014-12-30
test_hybrid.cpp: Fixing buffer overrun that was failing clang -fsanitize...

basking2:
test_hybrid.cpp: Fixing buffer overrun that was failing clang -fsanitize=address checks.


glongo:
Merged after rebase, thanks!

PR #89, opened 2015-02-18
Fixing small issues

glongo:
This fixes some warnings generated by strcat/sprintf functions
and also merges the following PR: [glongo#1] [#86]

Addresses the last PR: [https://github.com/OISF/libhtp/pull/88]

### DIFF
--- a/htp/htp_util.c
+++ b/htp/htp_util.c
@@ -2063,34 +2063,34 @@ void fprint_raw_data_ex(FILE *stream, const char *name, const void *_data, size_
     while (offset < len) {
         size_t i;
 
-        sprintf(buf, "%08" PRIx64, (uint64_t) offset);
-        strcat(buf + strlen(buf), "  ");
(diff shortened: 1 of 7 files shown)


htp/htp_util.c
```diff
@@ -2063,34 +2063,34 @@ void fprint_raw_data_ex(FILE *stream, const char *name, const void *_data, size_
     while (offset < len) {
         size_t i;
 
-        sprintf(buf, "%08" PRIx64, (uint64_t) offset);
-        strcat(buf + strlen(buf), "  ");
+        snprintf(buf, sizeof(buf), "%08" PRIx64, (uint64_t) offset);
+        strlcat(buf + strlen(buf), "  ", sizeof(buf));
```
inliniac: This is wrong: by using buf+strlen(buf) and then sizeof(buf) you are allowing a buffer overrun. If you look at my original comment you see how it should be.

PR #91, opened 2015-02-18
Fixing warnings

glongo:
This fixes the following warnings:
strlcpy.c:49: warning: function declaration isn't a prototype
strlcat.c:51: warning: function declaration isn't a prototype
htp_transcoder.c:163: warning: dereferencing type-punned pointer will break strict-aliasing rules


inliniac:
Looks good, lets merge it.

Issue #99, opened 2015-02-25
Problems with including config.h in htp_private.h


b1v1r:
79c2e9e2decf41fbebdc4d99fb6c485b9974e2bd broke being able to include htp_private.h in another product. Yes, we should not be doing this, but there is not a way to look at htp_cfg_t otherwise. This structure was made private, but there are no accessors (only setters) that are public.

It is a really bad idea to include the autoconf header in another header as the config.h has no include guards and uses too many generic names that will conflict (e.g., VERSION). The config.h is mean to ONLY be included from a source file and not another header. 

Please consider reverting 79c2e9e2decf41fbebdc4d99fb6c485b9974e2bd. I have done so in ironbee codebase until other issues surrounding private structures can be resolved.

catenacyber:
Looks fixed by #101

PR #100, opened 2015-02-25
Make including the autoconf config header safer, resolving issue #99.

b1v1r:
Here is a pull request to resolve issue #99 dealing with autoconf header.


b1v1r:
Missed some files in this. Will open another.

PR #105, opened 2015-06-16
htp/htp_util.c: Do not invoke callbacks in htp_req_run_hook_body_data…

basking2:
…() when there is no tx running.

This situation was encountered when load testing an Apache Traffic Server
plugin that produced responses, not from the Origin Server but, directly from the
plugin.


b1v1r:
Should be pulled into 0.5.x as well (not sure master is used).

glongo:
Could be possible to have a test case?

basking2:
I'm not sure I understand what went on in the lab simulation well enough to make a test that reflects what we saw. I would just be contriving my own invention of how to pass in a NULL in_tx value.

glongo:
Merged, thanks!

PR #108, opened 2015-07-09
[preview] Multi content encoding support v6

inliniac:
Support multiple layers of compression in response bodies. 2 layers seem
quite common in some proxy environments.

To do so, make htp_tx_t::htp_decompressor_t a list instead of a single
instance. This allows for stacked encoding.

Setting up of multiple content encoding is a slow path.

Add a limit to how many layers of compressed data will be uncompressed.
Defaults to 2, but is configurable. Setting it to 0 disables the limit.

If the limit is reached, a warning will be recorded.

### DIFF
--- a/htp/htp_config.c
+++ b/htp/htp_config.c
@@ -156,8 +156,9 @@ htp_cfg_t *htp_config_create(void) {
     cfg->parse_request_cookies = 1;
     cfg->parse_request_auth = 1;
     cfg->extract_request_files = 0;
-    cfg->extract_request_files_limit = -1; // Use the parser default.   
-        
+    cfg->extract_request_files_limit = -1; // Use the parser default.
+    cfg->response_decompression_layer_limit = 2; // 2 layers seem fairly common
+
     // Default settings for URL-encoded data.
 
     htp_config_set_bestfit_map(cfg, HTP_DECODER_DEFAULTS, bestfit_1252);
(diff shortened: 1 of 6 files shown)


inliniac:
@wxsBSD @b1v1r your feedback would be appreciated!

wxsBSD:
I can't really dive into this for a while as I'm unable to use anything but my mobile for another week or so. This sounds conceptually similar to the multiple content length bug I fixed a while back. If major browsers will handle multiple content encodings transparently then libhtp should do the same.

Do you have an example PCAP of this you could send? Or maybe a test for it. One thing that is not clear to me is if a server sends multiple content encodings (eg gzip twice) but only does a single encoding what do browsers do? Do they silently ignore the second encoding?

As I said, I won't be able to dig into this deeply for at least a week and I only quickly glanced at the code, but I believe libhtp should try to mimic major browser behavior and this sounds like a good candidate for that!

inliniac:
I don't have a pcap I can share, but it's not hard to generate one using: http://noxxi.de/research/content-encoding-online-scanner.html

I will be updating the support later with heuristic detection. (Some) browsers will apparently just figure out how to decompress even if the CE setting is wrong (i.e. claims to be deflate, is gzip). Thats currently not handled in my PR.

However, in the current code just having 2 gzip headers present will break decompression completely. So as a first step, it makes sense to support that case. Mixes of gzip and deflate are also supported in the PR.

Wrt claiming double gzip, but just one present: this will result in a decompress error being recorded in the stream, which we can match on in Suricata. When looking at the heuristic approach later I can see if I could handle this case as well.

wxsBSD:
Ok. I think that all makes sense. As long as libhtp behaves like major browsers in these situations in happy, in the cases where it isn't possible or not worth the effort then recording an error to be properly handled makes sense.

Thanks for the pointer to generating it myself, I'll play with it a bit when I can get back to my laptop.

inliniac:
Development continued in #139

PR #110, opened 2015-07-14
Connect plain http v8

inliniac:
In some cases, the tunneled data in a CONNECT tunnel is actually regular HTTP. In this case the parsing of HTTP can and should continue.

This patch implements this by adding a stage to the CONNECT handling. The first data of the tunnel is probed to see if it looks like HTTP. If it does, the new data is parsed and the new HTTP transaction is tracked normally.

If the probing fails to find HTTP, the old behavior is being used.

Changes since #109:
- set proper response state on connect tunnel confirm


inliniac:
Replaced by #111

PR #113, opened 2015-09-14
fix pkgconfig include dir

regit:
Software using libhtp are including 'htp/htp.h' so the path must
simply be ${includedir} and not ${includedir}/htp.


glongo:
Merged, thanks Eric!

PR #117, opened 2015-09-21
test/test.c: Coverity-identified issue of not checking malloc() retur…

basking2:
test/test.c: Coverity-identified issue of not checking malloc() return value.


glongo:
Merged, thanks!

PR #118, opened 2015-09-21
    test/test.c: Coverity-identified missing break in switch.

basking2:
test/test.c: Coverity-identified missing break in switch.


glongo:
Merged, thanks!

PR #119, opened 2015-09-22
Fix uninitialized htp_tx_t::is_last value in htp_tx_res_process_body_…

montekki:
…data_ex()

This value should obviously be initialized otherwise this leads to first TX data chunk captured being is_last since the value of d allocated on stack is uninitialized and != 0.


glongo:
Merged, thanks!

PR #120, opened 2015-09-24
Trigger request line parsing on incomplete request

inliniac:
When a response is received before the full request is received, we
are handling either with timeout case (where the server no longer
waits for the rest of the request) or a request too big case.

In such cases the request line would not be parsed if the state for
the request was still 'request line'.

This patch forces request line parsing in such cases.


glongo:
Merged, thanks!

PR #122, opened 2015-09-28
Fix uninitialized htp_tx_t::is_last value in htp_decompressors.c

montekki:
Fixed some more uninitialized values found while testing file extraction. Not sure that's all of them, all htp_tx_data_t local variables allocated on stack seem to have this issue. Please review.


glongo:
Merged, thanks!

PR #125, opened 2015-10-23
connect: handle response headers with 200 response

inliniac:
Lots of CONNECT requests are answered by:
    HTTP/1.0 200 Connection Established\r\n\r\n

The 2nd set of \r\n is considered to be part of the header. Due to some
missing header handling, in the CONNECT case this would be handled after
the tx was cleaned up. This would lead to the invocation of the
response header callback with a NULL tx.


glongo:
Merged, thanks!

PR #128, opened 2016-01-31
Accept responses without request

felix-jia:
This change is to support the async-oneside mode in Suricata. Under this mode, it is possible that responses do not match to any request, but these responses need to be inspected by Suricata.

This also answers the bug report for https://redmine.openinfosecfoundation.org/issues/993

### DIFF
--- a/htp/htp_response.c
+++ b/htp/htp_response.c
@@ -865,6 +865,12 @@ htp_status_t htp_connp_RES_FINALIZE(htp_connp_t *connp) {
     return htp_tx_state_response_complete_ex(connp->out_tx, 0 /* not hybrid mode */);
 }
 
+static inline long long _get_current_time_in_seconds() {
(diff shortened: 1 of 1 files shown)


inliniac:
Wonder if it would make sense to use a more explicit URI. E.g. "/libhtp::request_uri_not_seen"
@b1v1r @wxsBSD any thoughts?

wxsBSD:
I'm in favor of a more specific URI for sure. You can avoid the ugly snprintf with a static buffer by doing so. Why go with the timestamp?

felix-jia:
Made the change to use a static URI.

Originally, was trying to differentiate these responses using timestamp for logging and debugging purposes. A static URI simplifies the code.

victorjulien:
Merged this as part of #163, thanks a lot!

PR #129, opened 2016-02-16
headers: fix memory leak on malformed headers

inliniac:
Some malformed traffic lead to temporary header buffers not getting
freed at connection parsers destruction.


glongo:
Merged, thanks!

PR #140, opened 2016-07-05
Fixed 'response_decompression_enabled' being ignored in case of multiple encodings

WGH-:


inliniac:
Good catch, thank you!

PR #142, opened 2016-07-25
Fix warning on OS X.

wxsBSD:
When building on OS X (haven't tried others) there is a warning when building htp_util.c.

```
htp_util.c:2088:79: warning: format specifies type 'void *' but the argument has type 'const unsigned char *' [-Wformat-pedantic]
    fprintf(stream, "\n%s: ptr %p offset %" PRIu64 " len %" PRIu64"\n", name, data, (uint64_t) offset, (uint64_t) len);
                               ~~                                             ^~~~
                               %s
1 warning generated.
```

This commit fixes it by casting to a void pointer.


inliniac:
Merged through #144, thanks Wesley!

Issue #141, opened 2016-07-13
Supporting 101 Switching Protocols


WGH-:
Related to #13 which was closed for no apparent reason.

As soon as server replies with 101 Switching Protocols, anything after HTTP headers is no longer HTTP protocol.

> The server will switch protocols to those defined by the response's Upgrade header field immediately after the empty line which terminates the 101 response.

libhtp seems to ignore this status code, still trying to parse everything that comes after as HTTP data.

This leads to WebSocket traffic being interpreted as some junk requests/responses (I will attach some pcaps a bit later, if needed).

I believe the correct solution would be to stop tracking HTTP connection after 101 response, unless specific protocol can be actually supported.

WGH-:
I've started working on implementing this. As libhtp already does something similar with CONNECT requests, it shouldn't be very hard.

catenacyber:
Is this still an issue ? (I do not think so)

PR #143, opened 2016-08-04
on "101 Switching Protocols", treat connection as a tunnel #141

WGH-:
This fixes WebSockets traffic (WebSockets being the main user of HTTP Upgrade) being interpreted as garbage HTTP traffic.


inliniac:
Merged, thanks!

PR #151, opened 2017-06-14
libhtp: delete library version from SONAME

aborrero:
Delete library version from SONAME and bump it from 1:0:0 to 2:0:0.
Using the library version in the SONAME means that the effective SONAME is
bumped each release of libhtp, which have a negative impact in the integration
of the library in downstream distros.

For example, in Debian, the SONAME is put into the library package name to
allow co-existence of different API/ABI versions and to ease upgrade
management between versions.

The bump from 1 to 2 is made in order to left behind the old scheme.

Before this patch:

% objdump -p htp/.libs/libhtp.so | grep SONAME
  SONAME               libhtp-0.5.24.so.1

After this patch:

% objdump -p htp/.libs/libhtp.so | grep SONAME
  SONAME               libhtp.so.2

Signed-off-by: Arturo Borrero Gonzalez <arturo@debian.org>


inliniac:
Merged this in #155, thanks Arturo!

PR #152, opened 2017-06-20
docs/: cleanup directory

aborrero:
Drop sources that aren't included in the suricata release tarball.
The most important removal is the docbook-xsl-ns-1.77.1 source, which
clearly doesn't belong here.

Signed-off-by: Arturo Borrero Gonzalez <arturo@debian.org>


inliniac:
Why does this remove docs/COMPATIBILITY_CHANGES as well?

aborrero:
> Why does this remove docs/COMPATIBILITY_CHANGES as well?

because it isn't included in the suricata tarball

inliniac:
I think this file should be added to the dist then. Looks like other files in docs/ are in EXTRA_DIST in the 'root' Makefile.am

aborrero:
Ok, resending PR

PR #161, opened 2017-12-01
Clean response line when processed as body

bolzzy:
Clean response line when processed as body.
[Bug #2260](https://redmine.openinfosecfoundation.org/issues/2260) / [160](https://github.com/OISF/libhtp/issues/160)


inliniac:
Merged through #162, thanks!

PR #163, opened 2018-01-30
Feature/missing request/v8

victorjulien:
Support picking up a connection if it starts with the response.


victorjulien:
Based on and includes #128

Issue #106, opened 2015-07-07
configure doesn't honour --disable-debug and builds with -DHTP_DEBUG


barn:
Which causes suricata to output all the debug information to stdout, like in http://pastebin.com/YSHzyxNT

On OSX, 10.9. Trying to build for homebrew, but just git clone-ing and building directly gets these results.

```
bash-3.2# uname -v
Darwin Kernel Version 12.6.0: Wed Dec 17 19:11:40 PST 2014; root:xnu-2050.48.15~1/RELEASE_X86_64
bash-3.2# sh autogen.sh
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal --force -I m4
autoreconf: configure.ac: tracing
autoreconf: running: glibtoolize --copy --force
glibtoolize: putting auxiliary files in '.'.
glibtoolize: copying file './ltmain.sh'
glibtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'm4'.
glibtoolize: copying file 'm4/libtool.m4'
glibtoolize: copying file 'm4/ltoptions.m4'
glibtoolize: copying file 'm4/ltsugar.m4'
glibtoolize: copying file 'm4/ltversion.m4'
glibtoolize: copying file 'm4/lt~obsolete.m4'
autoreconf: running: /usr/local/Cellar/autoconf/2.69/bin/autoconf --force
autoreconf: running: /usr/local/Cellar/autoconf/2.69/bin/autoheader --force
autoreconf: running: automake --add-missing --copy --force-missing
configure.ac:79: installing './compile'
configure.ac:82: installing './config.guess'
configure.ac:82: installing './config.sub'
configure.ac:7: installing './install-sh'
configure.ac:7: installing './missing'
Makefile.am: installing './INSTALL'
htp/Makefile.am: installing './depcomp'
parallel-tests: installing './test-driver'
autoreconf: Leaving directory `.'
bash-3.2# ./configure  --prefix=/usr/local/Cellar/suricata/2.0.8 --disable-debug
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... ./install-sh -c -d
checking for gawk... no
checking for mawk... no
checking for nawk... no
checking for awk... awk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
Debug mode enabled
```

[...]

PR #165, opened 2018-04-13
Fix for impossible-to-disable configure flags

smithjacobj:
The --enable-debug, --enable-devmode, and --enable-gcov flags were all
set to force the value of "yes" when the flag was set, _not_ when not
set, due to a missing AC_ARG_ENABLE comma.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -51,7 +51,7 @@ dnl --------------------------------
 dnl Options
 dnl --------------------------------
 
-AC_ARG_ENABLE(debug, [  --enable-debug  Enable debug mode], [ enable_debug=yes ])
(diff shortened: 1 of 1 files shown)


victorjulien:
Now, in the 'if-not-set' case, they all default to 'yes', while they all should default to 'no'. Coincidentally, this is why clang fails, as it doesn't like one of the options (my bet is on gcov)

smithjacobj:
@victorjulien OK - it seemed like they wanted to default to "yes" - but I can definitely set the default to "no" instead.

smithjacobj:
I'm going to discuss this in #106 then submit a new PR with the desired behavior.

victorjulien:
I have a patch ready for merge

PR #175, opened 2018-12-19
Next/20181219/v3

victorjulien:
#174, plus:
- fix undefined behavior: signed int overflow
- fix compile warnings in tests
- improve status code parsing

cc @catenacyber


catenacyber:
Looks good to me.

There is some code that may be removed in `bstr_util_mem_to_pint` 
Once we checked for the overflow, we do not have to do it again twice...

victorjulien:
Agreed. Simplified it in #176

PR #177, opened 2019-01-08
Optimizes htp_list_array_replace avoiding useless loop

catenacyber:
Going from 7 to 6 seconds on fuzzing corpus

I am not sure that we need to check for the integer overflow...

### DIFF
--- a/htp/htp_list.c
+++ b/htp/htp_list.c
@@ -172,18 +172,7 @@ htp_status_t htp_list_array_replace(htp_list_array_t *l, size_t idx, void *e) {
 
     if (idx + 1 > l->current_size) return HTP_DECLINED;
 
-    size_t i;
-    // Do we need to check for this integer overflow ?
-    if (l->first > SIZE_MAX - idx) {
-        // l->first + idx > SIZE_MAX =>
-        // l->first + idx > l->max_size =>
-        // idx > l->max_size - l->first
-        i = idx - (l->max_size - l->first);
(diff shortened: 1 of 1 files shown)


catenacyber:
By the way, this solves the timeout referenced in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=12354

htp/htp_list.c
```diff
@@ -172,12 +172,15 @@ htp_status_t htp_list_array_replace(htp_list_array_t *l, size_t idx, void *e) {
...
     if (idx + 1 > l->current_size) return HTP_DECLINED;
 
-    size_t i = l->first;
-
-    while (idx--) {
-        if (++i == l->max_size) {
-            i = 0;
-        }
+    size_t i;
+    // Do we need to check for this integer overflow ?
```
victorjulien: not sure actually, size_t is 32bit on 32bit machines. To get to 4 billion transactions (the only place this code is used) we'd need a lot of traffic. But then again it's probably not completely impossible. From a quick look at the rest of the list code I don't get the impression a overflow is really handled in other places. Would you agree?

catenacyber: Indeed, it is not checked in `htp_list_array_get` where there is the same addition `l->first + idx`

victorjulien:
Can you do a new PR a single commit? I don't think a PR should have commits fixing other commits in the same PR. Thanks!

PR #178, opened 2019-01-08
Adds travis builds with sanitizers support

catenacyber:
We could then add the whole generated fuzzing corpus to be tested (to test for regressions with new pull requests for instance)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ before_install:
 matrix:
     include:
         - name: fuzza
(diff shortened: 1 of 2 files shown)


victorjulien:
Any idea on why the builds fail in travis?

catenacyber:
Not really, I copied this from https://github.com/aquynh/capstone/blob/master/.travis.yml which I wrote

The problem seems to be with the use of c++ in the test directory
I will look into it later

catenacyber:
It looks like travis is not using clang as a C++ compiler event though there is the compiler:clang option...

catenacyber:
Ok, so indeed, it is because travis did not use clang++ as C++ compiler. This can at least be fixed by setting CXX environment variable

catenacyber:
So here is a fix. I do not know if you have a better one so that travis and autotool configure agree on having clang++ as c++ compiler

victorjulien:
Can you do a new PR with all commits squashed? Thanks!

PR #180, opened 2019-02-01
More complete fuzzing with all suricata callbacks

catenacyber:
As discussed, @victorjulien , this is intended to get a better code coverage

There will then remain dead code to be removed.
But we will get a better view of which code is dead this way.

### DIFF
--- a/test/fuzz/fuzz_htp.c
+++ b/test/fuzz/fuzz_htp.c
@@ -91,7 +91,7 @@ static int HTPCallbackResponseStart(htp_tx_t *tx)
 
 static int HTPCallbackRequestLine(htp_tx_t *tx)
 {
-    fprintf(logfile, "HTPCallbackRequestLine %s\n", tx->parsed_uri);
+    fprintf(logfile, "HTPCallbackRequestLine\n");
     return 0;
 }
 
--- a/test/fuzz/fuzz_htp.c
+++ b/test/fuzz/fuzz_htp.c
@@ -26,12 +26,72 @@ FILE * logfile = NULL;
  *
  * @param[in] connp
  */
-static int callback_response(htp_tx_t *out_tx) {
+static int HTPCallbackResponse(htp_tx_t *out_tx) {
     if (out_tx != NULL) {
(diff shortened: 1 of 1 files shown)


test/fuzz/fuzz_htp.c
```diff
@@ -26,22 +26,82 @@ FILE * logfile = NULL;
...
+
+static int HTPCallbackResponseStart(htp_tx_t *tx)
+{
+    fprintf(logfile, "HTPCallbackResponseStart\n");
+    return 0;
+}
+
+static int HTPCallbackRequestLine(htp_tx_t *tx)
+{
+    fprintf(logfile, "HTPCallbackRequestLine %s\n", tx->parsed_uri);
```
victorjulien: tx->parsed_uri is a pointer to a struct of type htp_uri_t, not a printable string. In general most buffers in htp are byte arrays, not C strings.

catenacyber: I removed it to keep things simple

victorjulien:
Merged as part of #184. I did rewrite the commit message to follow Suricata's style.

victorjulien:
Ok merging this was a mistake as Suricata's QA is now broken. See https://travis-ci.org/OISF/suricata/jobs/495085015

Can you add some travis checks that would have caught this?

catenacyber:
Where can I see that Suricata's QA ?
The link you pointer shows green...

victorjulien:
On 19-02-19 08:46, Catena cyber wrote:
> Where can I see that Suricata's QA ?
> The link you pointer shows green...

I already pushed fixes, so I don't think you can see it now.

catenacyber:
Ok, I just saw your commits, thanks.
Where can I check next time ?

victorjulien:
On 19-02-19 10:00, Catena cyber wrote:
> Ok, I just saw your commits, thanks.
> Where can I check next time ?

In the OISF/suricata travis, but really this shouldn't happen again. I
would like to make sure the libhtp travis catches this so it doesn't
accidentally get merged.

Suricata bundles libhtp and then uses:
https://github.com/OISF/suricata/blob/master/.travis.yml#L43

PR #181, opened 2019-02-12
Optimizes htp_list_array_replace avoiding useless loop

catenacyber:
Going from 7 to 6 seconds on fuzzing corpus

Following #177 in one single squashed commit


victorjulien:
Merged as part of #184. I did rewrite the commit message to follow Suricata's style.

PR #186, opened 2019-02-22
headers: limits to two the number of duplicate headers

catenacyber:
This comes from a timeout generated by oss-fuzz
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=13255

We can abuse libhtp to do lots of `realloc` with headers such as :
```
a:1
a:2
a:3
a:1
```
that get transformed into `a:"1, 2, 3, 1"`

See discussion here : https://github.com/google/oss-fuzz/issues/2180

I have two points for which I would like an opinion :

@victorjulien when you said 
> raise the event
do you mean a call to `htp_log` or some other mechanism ?

Is setting the `HTP_FIELD_FOLDED` flag for the header the right way to limit the number of raised events ?

### DIFF
--- a/htp/htp_response_generic.c
+++ b/htp/htp_response_generic.c
@@ -256,16 +256,17 @@ htp_status_t htp_process_response_header_generic(htp_connp_t *connp, unsigned ch
     htp_header_t *h_existing = htp_table_get(connp->out_tx->response_headers, h->name);
     if (h_existing != NULL) {
         // Keep track of repeated same-name headers.
-        if (h_existing->flags & HTP_FIELD_REPEATED) {
-            // This is at least the third occurence for this header.
+        if (h_existing->flags & HTP_FIELD_REPEATED == 0) {
+            // This is the second occurence for this header.
             htp_log(connp, HTP_LOG_MARK, HTP_LOG_WARNING, 0, "Repetition for header");
+        } else {
             bstr_free(h->name);
             bstr_free(h->value);
             free(h);
             return HTP_OK;
         }
         h_existing->flags |= HTP_FIELD_REPEATED;
-                
+
         // Having multiple C-L headers is against the RFC but many
         // browsers ignore the subsequent headers if the values are the same.
         if (bstr_cmp_c_nocase(h->name, "Content-Length") == 0) {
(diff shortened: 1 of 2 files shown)


htp/htp_response_generic.c
```diff
@@ -256,6 +256,17 @@ htp_status_t htp_process_response_header_generic(htp_connp_t *connp, unsigned ch
     htp_header_t *h_existing = htp_table_get(connp->out_tx->response_headers, h->name);
     if (h_existing != NULL) {
         // Keep track of repeated same-name headers.
+        if (h_existing->flags & HTP_FIELD_REPEATED) {
+            // This is the third occurence for this header.
+            if (!(h_existing->flags & HTP_FIELD_FOLDED)) {
```
victorjulien: I don't think this folded flag should be used. Can we just rely on the HTP_FIELD_REPEATED flag? I think it's okay to allow just a single duplicate and that this should come with the warning already.

catenacyber: Ok, the downside is that we will get 998 calls to `htp_log` if we have 1000 duplicate headers

htp/htp_response_generic.c
```diff
@@ -256,6 +256,14 @@ htp_status_t htp_process_response_header_generic(htp_connp_t *connp, unsigned ch
     htp_header_t *h_existing = htp_table_get(connp->out_tx->response_headers, h->name);
     if (h_existing != NULL) {
         // Keep track of repeated same-name headers.
+        if (h_existing->flags & HTP_FIELD_REPEATED) {
+            // This is at least the third occurence for this header.
+            htp_log(connp, HTP_LOG_MARK, HTP_LOG_WARNING, 0, "Repetition for header");
```
victorjulien, line 261: if we only call this if is not set HTP_FIELD_REPEATED, the we'd just call it once right?
```c
if (h_existing != NULL) {
    if (h_existing->flags & HTP_FIELD_REPEATED == 0) {
        htp_log(connp, HTP_LOG_MARK, HTP_LOG_WARNING, 0, "Repetition for header");
    } else {
        bstr_free(h->name);
        bstr_free(h->value);
        free(h);
        return HTP_OK;
    }
}

catenacyber: Yes, but this will issue a warning starting from the first repetition. Isn't it too much ?

victorjulien: I think warning on any duplicate header is already useful. I don't think it should really happen in normal traffic. Or do you think I'm missing something?

catenacyber: I saw it happen on "normal traffic", some duplicate secure header like `X-XSS-Protection:1` to be really sure it is secure :-)
So yes, this was misconfiguration, and it was worth reporting...

catenacyber:
I can squash all the commits when the pull request is good :-)

catenacyber:
Continues in #188

PR #187, opened 2019-03-04
http: handles multiple content-length headers

catenacyber:
Logs only as warning
Takes the latest value as the valid one
Same goes for requests and responses

Fixes #1776

I assume that Suricata should parse as much as possible.
So, we should not return HTP_ERROR to keep parsing.
As Wireshark, we use the latest header as the valid one.

Is this assumption right ?
If it is the case, should I create test for suricate-verify ?

### DIFF
--- a/htp/htp_request_generic.c
+++ b/htp/htp_request_generic.c
@@ -87,9 +87,9 @@ htp_status_t htp_process_request_header_generic(htp_connp_t *connp, unsigned cha
                 bstr *tmp = h_existing->value;
                 h_existing->value = h->value;
                 h->value = tmp;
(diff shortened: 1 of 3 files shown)


victorjulien:
Yeah, can you create a Suricata PR if needed for the new warning and a suricata-verify PR for testing various combinations of requests and responses with CL issues?

catenacyber:
There are already sample tests in libhtp for C-L issues : 74-response-multiple-cl.t  and 88-response-multiple-cl-mismatch.t

But I should still add some tests to suricata-verify, right ?

catenacyber:
See #189 squashed

PR #188, opened 2019-03-04

catenacyber:
New version of #186


victorjulien:
We've found an issue with this, where seemingly benign traffic has lots of duplicate headers. I've sent a test case privately.

PR #189, opened 2019-03-04
http: handles multiple content-length headers

catenacyber:
Logs only as warning
Takes the latest value as the valid one
Same goes for requests and responses

Fixes #1776
Follows #187


catenacyber:
Closing in favor of #194

PR #190, opened 2019-03-05
Returns error on invalid request line

catenacyber:
Fixes #2655

We could be more strict and accept only `GET` methods for HTTP 0.9

Test case 42 had to be rewritten as it is not valid HTTP 0.9


victorjulien:
I'm wondering if we should be more strict. I agree the warning should be raised, but in general we try to be liberal as we value getting something out of the stream more than being strictly correct in most cases. Experience shows that many clients and servers are also much more relax and forgiving than RFCs would allow.

catenacyber:
> we try to be liberal as we value getting something out of the stream more than being strictly correct in most cases

I totally agree and that is what I am trying to do here.
This patch is not meant to be RFC compliant, it is about acknowledging that we lost track.
But I surely do not see which are the implications for having `HTPHandleRequestData` return error.
I thought it would restart parsing from a fresh state on next packet.
By your response, I guess I thought wrong and parsing does not resume.
Can you tell ?

I guess that you mean that we should consider anything between the end of a request and the beginning of a new request to be `http_client_body`.
I will start to do that.
Best would be to have a new keyword `http_client_junk` but that can be done afterwards.
Anyways, that will just cover the case in the pcap and bypasses will remain.

victorjulien:
I'm not entirely sure, but I think if we return error ultimately we go into an error state in Suricata and stop feeding the parser new data. But this should be easy enough to test with a unittest in Suricata.

catenacyber:
> we go into an error state in Suricata and stop feeding the parser new data

This is not what I intended indeed

PR #191, opened 2019-03-05
Returns error on invalid request line

catenacyber:
Fixes #2655

Follows #190

PR #192, opened 2019-03-06
Adds test cases for compression

catenacyber:
Wrong method is advertised
Multiple chained methods

Fixes #941

Test case 91 is in PR #191


victorjulien:
The tests appear to fail in travis.

catenacyber:
Yes, this was a dummy git mistake

PR #193, opened 2019-03-10
fix build without GNU libiconv

ffontaine:
iconvctl is only defined in GNU libiconv so check for the availability
of this function before using it

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


catenacyber:
Looks good to me

victorjulien:
Merged in #293, thanks for your patience!

PR #194, opened 2019-04-02
Http evader v1

catenacyber:
Fixes all http evader cases


catenacyber:
I tried to make each commit a logical unit.
If you want to see my original work, you can check https://github.com/catenacyber/libhtp/tree/http-evader-v0

I guess this needs some review.
Here are some points already
- should LZMA decompression support be a compile time option ?
- any better idea for elegant code and the different end of lines handling ?

catenacyber:
Code for LZMA is in public domain and comes from 
https://www.7-zip.org/sdk.html

htp/htp_response.c
```diff
@@ -550,21 +550,29 @@ htp_status_t htp_connp_RES_BODY_DETERMINE(htp_connp_t *connp) {
         }
     }
 
+    htp_header_t *ct = htp_table_get_c(connp->out_tx->response_headers, "content-type");
```
victorjulien, line 553: ct doesn't look like it's used before the old location of invocation, so it's probably best to move it back? These table lookups are not free, so best to only do them as needed.

htp/htp_decompressors.c
```diff
@@ -207,6 +207,8 @@ static htp_status_t htp_gzip_decompressor_decompress(htp_decompressor_gzip_t *dr
         dout.tx = d->tx;
         dout.data = NULL;
         dout.len = 0;
+        // We should be called for a connection end and do
```
victorjulien, line 210: I don't understand this comment

htp/htp_util.c
```diff
@@ -251,7 +251,18 @@ int htp_is_line_whitespace(unsigned char *data, size_t len) {
...
+    size_t pos = 0;
+
+    if (len == 0) return -1003;
+
+    // Ignore junk before
+    while ((pos < len) && (data[pos] < '0' || data[pos] > '9')) pos++;
+    if (pos == len) return -1001;
+
+    return bstr_util_mem_to_pint(data + pos, len - pos, 10, &pos);
+    // Ok to have junk afterwards
```
victorjulien, line 265: could we set an event for this?

htp/htp_util.c
```diff
@@ -2366,14 +2366,17 @@ int htp_treat_response_line_as_body(const uint8_t *data, size_t len) {
     //      Firefox 3.5.x: (?i)^\s*http
     //      IE: (?i)^\s*http\s*/
     //      Safari: ^HTTP/\d+\.\d+\s+\d{3}
+    size_t pos = 0;
 
     if (data == NULL) return 1;
-    if (len < 4) return 1;
+    while ((pos < len) && (htp_is_space(data[pos]) || data[pos] == 0)) pos++;
```
victorjulien, line 2389: whats the logic of the test for the 0 byte?

htp/htp_response.c
```diff
@@ -830,10 +830,28 @@ htp_status_t htp_connp_RES_HEADERS(htp_connp_t *connp) {
...
+                    if (colon_pos < len &&
+                        bstr_chr(connp->out_header, ':') >= 0 &&
+                        connp->out_tx->response_protocol_number == HTP_PROTOCOL_1_1) {
+                        // Warn only once per transaction.
+                        if (!(connp->out_tx->flags & HTP_INVALID_FOLDING)) {
+                            connp->out_tx->flags |= HTP_INVALID_FOLDING;
+                            htp_log(connp, HTP_LOG_MARK, HTP_LOG_WARNING, 0, "Invalid response field folding");
+                        }
+                        if (connp->cfg->process_response_header(connp, bstr_ptr(connp->out_header),
+                            bstr_len(connp->out_header)) != HTP_OK) return HTP_ERROR;
```
victorjulien, line 891: could you move the return statements to their own line?

htp/htp_response.c
```diff
@@ -727,18 +727,50 @@ htp_status_t htp_connp_RES_BODY_DETERMINE(htp_connp_t *connp) {
  * @returns HTP_OK on state change, HTP_ERROR on error, or HTP_DATA when more data is needed.
  */
 htp_status_t htp_connp_RES_HEADERS(htp_connp_t *connp) {
+    int endwithCR;
```
victorjulien, line 739: please keep vars all lowercase

htp/htp_response_generic.c
```diff
@@ -59,7 +59,7 @@ htp_status_t htp_parse_response_line_generic(htp_connp_t *connp) {
     tx->response_message = NULL;
 
     // Ignore whitespace at the beginning of the line.
-    while ((pos < len) && (htp_is_space(data[pos]))) pos++;
+    while ((pos < len) && (htp_is_space(data[pos]) || data[pos] == 0)) pos++;
```
victorjulien, line 62: can you explain why this is needed and what the consequences are? Suricata has a 'http_raw_headers' keyword that expects raw data, so including unprintable chars I think

htp/htp_util.c
```diff
@@ -463,6 +463,12 @@ int htp_connp_is_line_terminator(htp_connp_t *connp, unsigned char *data, size_t
             if (htp_is_line_empty(data, len)) {
                 return 1;
             }
+            // Only space is terminator if terminator does not follow right away
+            if (len == 2 && htp_is_lws(data[0]) && data[1] == LF) {
+                if (connp->out_current_read_offset < connp->out_current_len && connp->out_current_data[connp->out_current_read_offset] != LF) {
```
victorjulien, line 468: style: line too long

htp/htp_response_generic.c
```diff
@@ -217,8 +219,52 @@ htp_status_t htp_parse_response_header_generic(htp_connp_t *connp, htp_header_t
...
+            nbnull++;
+        }
+    }
+    if (nbnull == 0) {
+        // simple case
+        h->name = bstr_dup_mem(data + name_start, name_end - name_start);
+    } else {
+        htp_log(connp, HTP_LOG_MARK, HTP_LOG_WARNING, 0,
+                "Null character in header line");
+        h->name = bstr_alloc(name_end - name_start - nbnull);
```
victorjulien, line 236: missing error check?

htp/htp_response_generic.c
```diff
@@ -217,8 +219,52 @@ htp_status_t htp_parse_response_header_generic(htp_connp_t *connp, htp_header_t
...
+
+    nbnull = 0;
+    j = 0;
+    for (i = value_start; i < value_end; i++) {
+        if (data[i] == 0) {
+            nbnull++;
+        }
+    }
+    if (nbnull == 0) {
+        h->value = bstr_dup_mem(data + value_start, value_end - value_start);
```
victorjulien, line 254: missing error check?

htp/htp_response_generic.c
```diff
@@ -217,8 +219,52 @@ htp_status_t htp_parse_response_header_generic(htp_connp_t *connp, htp_header_t
...
+    } else {
+        htp_log(connp, HTP_LOG_MARK, HTP_LOG_WARNING, 0,
+                "Null character in header line");
+        h->value = bstr_alloc(value_end - value_start - nbnull);
```
victorjulien, line 258: error check

victorjulien:
Seeing many build failures in my QA like this:
```
make[5]: Entering directory `/builds/inliniac/suricata-ci/suricata/libhtp/htp/lzma'
/bin/sh ../../libtool  --tag=CC   --mode=compile gcc -DHAVE_CONFIG_H -I. -I../..   -O2 -I../.. -D_GNU_SOURCE -g -Wall -Wextra -std=c11 -pedantic -Wextra -Wno-missing-field-initializers -Wshadow -Wpointer-arith -Wstrict-prototypes -Wmissing-prototypes -Wno-unused-parameter -O2 -Wstrict-overflow=1 -fstack-protector -D_FORTIFY_SOURCE=2 -Wformat -Wformat-security -fPIC -MT LzFind.lo -MD -MP -MF .deps/LzFind.Tpo -c -o LzFind.lo LzFind.c
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I../.. -O2 -I../.. -D_GNU_SOURCE -g -Wall -Wextra -std=c11 -pedantic -Wextra -Wno-missing-field-initializers -Wshadow -Wpointer-arith -Wstrict-prototypes -Wmissing-prototypes -Wno-unused-parameter -O2 -Wstrict-overflow=1 -fstack-protector -D_FORTIFY_SOURCE=2 -Wformat -Wformat-security -fPIC -MT LzFind.lo -MD -MP -MF .deps/LzFind.Tpo -c LzFind.c  -fPIC -DPIC -o .libs/LzFind.o
cc1: error: unrecognized command line option "-std=c11"
make[5]: *** [LzFind.lo] Error 1
```

victorjulien:
Another failure
```
Test HTPParserTest21                                              : /bin/bash: line 93: 16411 Segmentation fault      (core dumped) ./src/suricata -u -l /tmp/
ERROR: Job failed: exit code 1
```

htp/htp_response.c
catenacyber: Ok, well spotted

htp/htp_decompressors.c
catenacyber: This is indeed complicated.
So I will try to explain in a longer way.

The bigger question is do we have to output all the results we can with the data we get ? Or wait for more data to give a better result...
In this case, do we have to stream/output all the decompressed data (even if we did not reach content-length yet) as soon as we have it ? 
Or should we bufferize it and output it only when it is complete ?
In the latter case, `htp_gzip_decompressor_decompress` needs to be called in the case of a connection end so that it can still output the partially decompressed data.
But there is now no such mechanism in suricata/libhtp as to stop the connection but process the remaining buffered data.

This question comes again with other cases where we probe further data in order to interpret the current data, for instance https://github.com/OISF/libhtp/pull/195#issuecomment-478913730

Is this clearer Victor ?

htp/htp_util.c
catenacyber: Ok good idea

htp/htp_response.c
catenacyber: Ok, should I do it for the unmodified code just above ?

htp/htp_response.c
catenacyber: ok

htp/htp_util.c
catenacyber: I do not know if it is logic but, according to http-evader, some browsers allow space, tabs and null character before response line.
This is the case in http evader 390

htp/htp_util.c
catenacyber: Ok, split

htp/htp_response_generic.c
catenacyber: Ok, maybe there is a better code style for this whole thing...

htp/htp_response_generic.c
catenacyber: Ok

htp/htp_response_generic.c
catenacyber: Ok

htp/htp_response.c
victorjulien: No lets only do it for code you are modifying for other reasons.

htp/htp_response_generic.c
catenacyber: This is in fact not needed here, I remove it

A few lines below, the null characters check is needed for meaningful headers such as content-length.
When libhtp does `cl = htp_table_get_c(connp->out_tx->response_headers, "content-length")`
it does not work if it is `"content\x00-length"` instead of `"content-length"` but browsers do recognize them

So, is this ok ?

htp/htp_util.c
victorjulien: Ok thanks.

htp/htp_response_generic.c
victorjulien: Ok can you make sure the commit message clearly explains this? Also that it is about the header names, not values.

catenacyber: In fact, it is for both header names and header values.
As I remember, we have to do it for values for the case
`Transfer-encoding: chun\x00ked`

So, what is the expected behavior ?
Normalize names, but not values, and use a new `strcmp_with_zeroes` function for checking if transfer-encoding value is chunked ?

victorjulien: From Suricata's perspective, I would expect the following:

1. http_header keyword works on normalized keys and values, so nulls should be stripped
2. http_raw_header keyword should have the nulls in both keys and values
3. the nulls should lead to 'event' we can match on

victorjulien: I actually think that there should be 2 events: one for nulls in the keys and the other for nulls in the values. The former seems more serious to me.

catenacyber: There was actually already one log from libhtp for keys : "Response header name is not a token."
I added one for null characters for values.

PR #195, opened 2019-04-02
Try header parsing even if protocol is missing

catenacyber:
Fixes #2619


catenacyber:
This patch is not perfect as you can evade by splitting the packets, ie putting the headers in a packet after the request line.
But I wrote it this way, so as to have a decision made right away in case of real HTTP 0.9

PR #197, opened 2019-04-02

catenacyber:
Fixes #2881


victorjulien:
I see this PR causes a few matching failures when there is a request line that has spaces after the protocol, but before the end of line sequence.
```
GET / HTTP/1.0 \r\n
```
This leads to the proto becoming part of the URI. I'll share a test case privately.

Second issue is that one of the etpro sigs doesn't match anymore, as it expects the URI to end at the first space. Curiously wireshark agrees with it, and considers everything after the space to be part of the protocol. I'll share this test case privately as well.

catenacyber:
Thanks @victorjulien I fixed this

PR #198, opened 2019-04-03
Http evader v2

catenacyber:
Follows #194 with code review taken into account

PR #199, opened 2019-04-03
Http evader v3

catenacyber:
Follows #198 with rebase on 0.5.x


victorjulien:
For Suricata liblzma is an optional dependency. I wonder if we should do the same here, or make it mandatory on both sides.

victorjulien:
I see that you wondered about this in #194 as well. Lets make it optional for now.

catenacyber:
Indeed I have been wondering and forgot to mention it in the GitHub comments...
configure exits if it is not found, but compiled code works without it
I make it optional in http-evader-v4

PR #200, opened 2019-04-04
Http evader v4

catenacyber:
Follows #199 

- makes liblzma support optional (configure does not exit if it does not find it)
- rewords the commits to explain why
- reverts useless change in `htp_connp_is_line_ignorable`


catenacyber:
There is still to address the pending question from https://github.com/OISF/libhtp/pull/194#discussion_r272029858

PR #201, opened 2019-04-04
Http evader v5

catenacyber:
Follows #200 

Changed behavior regarding null characters in headers
These characters are now kept to be provided to Suricata
But libhtp uses new custom `strcmp` functions that skip zeroes when values are needed for further parsing, ie transfer-encoding is chunked or content-encoding indicates some kind of compression


configure.ac
```diff
@@ -109,6 +109,17 @@ if test "$ZLIB" = "no"; then
     exit 1
 fi
 
+# liblzma
+AC_CHECK_HEADER(lzma.h,AC_DEFINE([HAVE_LIBLZMA],[1],[liblzma available]),[AC_ERROR(lzma.h not found ...)])
```
victorjulien, line 113: Using AC_ERROR seems wrong here:
```
checking for lzma.h... no
configure: error: lzma.h not found ...
configure: error: ./configure failed for libhtp
ERROR: Job failed: exit code 1
```

victorjulien: I'm quickly dumped in the checks from suricata, but then I see the next issue: make check fails, presumably due to lzma dependencies there?

catenacyber: Thank you again Victor for the review

jmtaylor90: As an example, I think the lzma check in suricata configure.ac should work now in master as part of
https://github.com/OISF/suricata/commit/b98c28a60de49d71631ef81a273b88e1c7d3a028#diff-67e997bcfdac55191033d57a16d1408a

If you want to see if that format works in libhtp?

PR #202, opened 2019-04-05
Http evader v6

catenacyber:
Follows #201 

- configure does not fail without lzma.h (I had tested only adding #undef HAVE_LIBLZMA to the code)
- lzma test is not run if lzma is not found on the system
- adds a log message in the case of null characters in response header value (there is already one for non ascii-printable characters in header names)


htp/htp_response.c
```diff
@@ -558,25 +558,32 @@ htp_status_t htp_connp_RES_BODY_DETERMINE(htp_connp_t *connp) {
...
+
     // Check for "101 Switching Protocol" response.
     // If it's seen, it means that traffic after empty line following headers
     // is no longer HTTP. We can treat it similarly to CONNECT.
     // Unlike CONNECT, however, upgrades from HTTP to HTTP seem
     // rather unlikely, so don't try to probe tunnel for nested HTTP,
     // and switch to tunnel mode right away.
     if (connp->out_tx->response_status_number == 101) {
-        connp->out_state = htp_connp_RES_FINALIZE;
+        if (cl == NULL || te != NULL) {
```
victorjulien, line 571: I'm not sure I understand this line: if either there is no CL, or if there is a TE, then switch to htp_connp_RES_FINALIZE. Can you explain the CL/TE logic?

catenacyber: This is evader case 448
The 101 response code seems to become meaningless (for browsers) by the presence of content-length header (or transfer-encoding)
The logic is to follow the browsers logic even if does not fully make sense to us at first sight.
What do you think ?

(PS : I tried suricata-verify without this patch to find the evader test)

victorjulien: If I understand your explanation, the presence of either CL or TE should trigger this new handling. So shouldn't it be
```c
if (cl == NULL && te == NULL) {
    connp->out_state = htp_connp_RES_FINALIZE;
}
```

catenacyber: Correct, thanks

PR #203, opened 2019-04-18

catenacyber:
Fixes #2881

It is less likely that spaces are added in protocol than in URIs

Follows #197 :
- Allows whitespace between protocol and end of line

### DIFF
--- a/htp/htp_request_generic.c
+++ b/htp/htp_request_generic.c
@@ -331,25 +331,29 @@ htp_status_t htp_parse_request_line_generic_ex(htp_connp_t *connp, int nul_termi
 
     start = pos;
     bad_delim = 0;
+    pos = len - 1;
 
-    // The URI ends with the first whitespace.
-    while ((pos < len) && (data[pos] != 0x20)) {
+    // Skips the spaces at the end of line (after protocol)
+    while (pos > start && htp_is_space(data[pos])) pos--;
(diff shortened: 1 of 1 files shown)


htp/htp_request_generic.c
```diff
@@ -331,25 +331,29 @@ htp_status_t htp_parse_request_line_generic_ex(htp_connp_t *connp, int nul_termi
...
         // implementations allow other delimiters, like tab or other
         // characters that isspace() accepts.
-        pos = start;
         while ((pos < len) && (!htp_is_space(data[pos]))) pos++;
     }
     if (bad_delim) {
         // warn regardless if we've seen non-compliant chars
         htp_log(connp, HTP_LOG_MARK, HTP_LOG_WARNING, 0, "Request line: URI contains non-compliant delimiter");
+    } else if (pos == start) {
+            pos = len;
```
victorjulien, line 356: style: indent is off

victorjulien (commented):
I think I would want this behavior to be optional. Suricata 4.1.x should not change, but in Suricata 5.x I would like it to be the default. This I think means we'd have to be able to control the behavior from suri. Thoughts?

htp/htp_request_generic.c
catenacyber: ok

PR #204, opened 2019-04-18
Adds htp_parse_content_range function

catenacyber:
Fixes #2485

See https://github.com/OISF/suricata/pull/3773

This is a proposal.
Here are some open questions 
- Is libhtp the right place for this (or should we do this in suricata) ?
- Should a `htp_content_range_t` field be added to `htp_tx_t` ? (and be parsed automatically depending on configuration)
- How should we test it ?

### DIFF
--- a/htp/htp.h
+++ b/htp/htp.h
@@ -602,6 +602,32 @@ struct htp_uri_t {
 };
 
 /**
+ * Content-range structure. Where an element is not present, the
+ * corresponding field will be set to NULL or -1, depending on the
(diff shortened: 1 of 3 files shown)


victorjulien:
Updating htp_tx_t is tricky business, as it's meant to remain a abi that is backwards compat. Not impossible of course.

I think I'd prefer to move this into suri itself.

catenacyber:
Ok, so I am closing and moving this code to suricata

PR #206, opened 2019-05-02

catenacyber:

Follows #205 

Modified :
- use a counter in `htp_tx_t` instead of hacking `flags`

The counter is global per the transaction.
Ie if in one transaction, header a is repeated 30 times, and then header b is repeated 40 times, we will drop the last headers b even if 40 is less than 64...

### DIFF
--- a/htp/htp.h
+++ b/htp/htp.h
@@ -536,6 +536,9 @@ struct htp_tx_t {
 
     /** Transaction index on the connection. */
     size_t index;
+
+    /** Total repetitions for headers. */
+    uint32_t header_repetitions;
 };
 
 /**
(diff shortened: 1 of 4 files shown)


htp/htp.h
```diff
@@ -536,6 +536,9 @@ struct htp_tx_t {
 
     /** Transaction index on the connection. */
     size_t index;
+
+    /** Total repetitions for headers. */
+    uint32_t header_repetitions;
```
victorjulien, line 541: If I read this patch correctly, this field is shared between request and response. But I don't see it be reset anywhere?

catenacyber: Indeed, I thought `in_tx` and `out_tx` were different.

I tested with 
```
>>>
GET / HTTP/1.0
h: 112
h: 123
h: 124
h: 1r24
h: 1rzd24


<<<
HTTP/1.0 200 OK
Date: Mon, 31 Aug 2009 20:25:50 GMT
Server: Apache
Connection: close
Content-Type: text/html
Content-Length: 12
a:b
a:c
a:d
a:e
a:xd

Hello World!
```

PR #207, opened 2019-05-02

catenacyber:

Modifies #206
- Resets `header_repetitions` before moving to `htp_connp_RES_HEADERS`

### DIFF
--- a/htp/htp.h
+++ b/htp/htp.h
@@ -536,6 +536,9 @@ struct htp_tx_t {
 
     /** Transaction index on the connection. */
     size_t index;
+
+    /** Total repetitions for headers. */
+    uint32_t header_repetitions;
 };
 
 /**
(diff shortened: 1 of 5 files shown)


htp/htp.h
```diff
@@ -536,6 +536,9 @@ struct htp_tx_t {
 
     /** Transaction index on the connection. */
     size_t index;
+
+    /** Total repetitions for headers. */
+    uint32_t header_repetitions;
```
victorjulien, line 541: I think I would prefer splitting this into 2: request_header_repetitions and response_header_repetitions. Both could be uint16_t. The benefit I see is that no reset would be needed.

victorjulien: To clarify, I'd like to avoid the reset as the flow of the htp state machine is quite obscure even to me as a dev. So keeping things clean would make things easier to follow/debug.

catenacyber: You are right :-)

PR #209, opened 2019-05-02

catenacyber:

Modifies #207 :  
- Splits `header_repetitions` in two : `req_header_repetitions` and `res_header_repetitions`
and no longer resets it before response headers

### DIFF
--- a/htp/htp_response_generic.c
+++ b/htp/htp_response_generic.c
@@ -256,8 +256,22 @@ htp_status_t htp_process_response_header_generic(htp_connp_t *connp, unsigned ch
     htp_header_t *h_existing = htp_table_get(connp->out_tx->response_headers, h->name);
     if (h_existing != NULL) {
         // Keep track of repeated same-name headers.
+        if ((h_existing->flags & HTP_FIELD_REPEATED) == 0) {
+            // This is the second occurence for this header.
+            htp_log(connp, HTP_LOG_MARK, HTP_LOG_WARNING, 0, "Repetition for header");
+        } else {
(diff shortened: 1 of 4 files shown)


htp/htp_response_generic.c
```diff
@@ -256,8 +256,22 @@ htp_status_t htp_process_response_header_generic(htp_connp_t *connp, unsigned ch
...
     if (h_existing != NULL) {
         // Keep track of repeated same-name headers.
+        if ((h_existing->flags & HTP_FIELD_REPEATED) == 0) {
+            // This is the second occurence for this header.
+            htp_log(connp, HTP_LOG_MARK, HTP_LOG_WARNING, 0, "Repetition for header");
+        } else {
+            // For simplicity reasons, we count the repetitions of all headers
+            if (connp->out_tx->res_header_repetitions < HTP_MAX_HEADERS_REPETITIONS) {
+                connp->out_tx->res_header_repetitions++;
+            } else {
```
victorjulien, line 266: Maybe it would be good to add another warning here. Something 'excessive response header repetitions'

catenacyber: Ok, should I create an event in Suricata for this warning ?

victorjulien: Yes... I think the rule for that one should be enabled by default. The one based on a single repetition probably not.

We'd need the same for the request side btw.

catenacyber: Ok for the suri event.

I had guessed so for the request side, thanks for pointing it out

PR #210, opened 2019-05-06

catenacyber:

Modifies #209 
Adds a log message  "Excessive header repetitions"


victorjulien:
Merged in #212, thanks!

PR #213, opened 2019-05-07
Http close headers 2969 v1

catenacyber:
Fixes #2969

Another solution would be to change Suricata detect-http-header.c so that 
`if (AppLayerParserGetStateProgress(IPPROTO_TCP, ALPROTO_HTTP, tx, flags) <= HTP_RESPONSE_HEADERS)` does not apply when connection is closing

The ticket is about response headers but the same problem is present for request headers (and fixed in another commit)


victorjulien:
This is now merged as of #234

catenacyber:
Thanks, we can now merge https://github.com/OISF/suricata-verify/pull/55

PR #217, opened 2019-05-14
Fuzzcallbacks v1

catenacyber:
First commit is about compiling easily libhtp (I can provide my CMakeLists.txt files if you wish)

Second commit is extending the fuzz targets so that the callbacks with buffers returned from libhtp will check if the buffers are valid (ie if we do not overflow).
@victorjulien this follows our mail about ASAN triggered suricata.

### DIFF
--- a/htp/htp_private.h
+++ b/htp/htp_private.h
@@ -60,6 +60,7 @@ extern "C" {
 #include <sys/types.h>
 #include <sys/stat.h>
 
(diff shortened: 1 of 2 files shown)


catenacyber:
Should I cherry-pick https://github.com/OISF/libhtp/pull/221/commits/f8ab1a681002fe5b9d5082d9ff67d06a95446167 in this pull request ? (and remove it from http-evader)

PR #218, opened 2019-05-14
htp: better format strings for int64_t

catenacyber:
Depending on the system and the compiler, we can have a warning about `%d` not be a good format string fit for `int64_t`
Let's use the macro `PRId64`
Thanks https://stackoverflow.com/questions/9225567/how-to-print-a-int64-t-type-in-c


victorjulien:
Merged in #222, thanks!

PR #219, opened 2019-05-14
libhtp: remove direct fprintf to stderr

catenacyber:
Found by oss-fuzz

Libhtp should not use directly fprintf in normal mode (but htp_log)
This is the only case in whole corpus


victorjulien:
Merged in #222, thanks!

PR #220, opened 2019-05-14
Fuzzperf v1

catenacyber:
These two commits come from profiling libhtp with corpus from oss-fuzz

Out of the 596112 logs produced, we have the top 3 :
- 22008 Invalid response line: invalid protocol
- 133877 Request line: non-compliant delimiter between Method and URI
- 353607 Request line: URI contains non-compliant delimiter

So, I propose to disable the first two and get rid of around 80% of the logs 
My profiling shows that we go from 7.5 seconds to 6.5 seconds with this change on the corpus

The other change is using directly `htp_list_t` in `htp_table_t` instead of a pointer to it (thus avoiding one alloc/free)
That is a gain of about 0.15 seconds for my profiling

### DIFF
--- a/htp/htp_table_private.h
+++ b/htp/htp_table_private.h
@@ -62,7 +62,7 @@ enum htp_table_alloc_t {
 
 struct htp_table_t {
     /** Table key and value pairs are stored in this list; name first, then value. */
-    htp_list_t *list;
+    htp_list_t list;
 
     /**
      * Key management strategy. Initially set to HTP_TABLE_KEYS_ALLOC_UKNOWN. The
--- a/htp/htp_list.c
+++ b/htp/htp_list.c
@@ -42,6 +42,22 @@
 
 // Array-backed list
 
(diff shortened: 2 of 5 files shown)


htp/htp_table_private.h
```diff
@@ -62,7 +62,7 @@ enum htp_table_alloc_t {
 
 struct htp_table_t {
     /** Table key and value pairs are stored in this list; name first, then value. */
-    htp_list_t *list;
+    htp_list_t list;
```
victorjulien, line 65: I think this is going to break the ABI?

catenacyber: I don't think so. This is in htp_table_private.h so this is not public (the public structure is an opaque pointer) and we still use the functions the same way such as `htp_table_create`

We do add some functions to the ABI `htp_list_init` and `htp_list_release` but that remains compatible with previous functions.

victorjulien: Agreed. I tested it as well.

victorjulien:
Merged in #225, thanks!

PR #224, opened 2019-05-21
Closes connection in fuzz target

catenacyber:
So that fuzzing reflects more the reality with different actions taken
on connection closing (such as outputing partially decompressed data)


victorjulien:
Merged in #225, thanks!

PR #230, opened 2019-06-03
deadcode: removes some unused functions

catenacyber:
These were found using coverage by oss-fuzz
https://storage.googleapis.com/oss-fuzz-coverage/libhtp/reports/20190526/linux/src/libhtp/htp/report.html

Next step may be to add fuzz targets to increase coverage (like using `HTP_SERVER_IIS_5_1`or such)

### DIFF
--- a/htp/bstr.c
+++ b/htp/bstr.c
@@ -51,14 +51,6 @@ bstr *bstr_alloc(size_t len) {
     return b;
 }
 
-bstr *bstr_add(bstr *destination, const bstr *source) {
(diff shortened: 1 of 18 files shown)


victorjulien:
In a library, how can you know public functions are unused? They may not be used by the lib itself or by Suricata, but that doesn't mean others don't use them either.

catenacyber:
> In a library, how can you know public functions are unused?

Indeed, I don't.
Do you know other uses ?

If we need to keep all public functions, we can close this PR...

catenacyber:
So, I guess I should close this

PR #196, opened 2019-04-02
Try header parsing even if protocol is missing

catenacyber:
Fixes #2619

Follows #195

PR #236, opened 2019-07-10
Try header parsing even if protocol is missing

catenacyber:
Fixes #2619

Follows #196 fixing warning by using same type


victorjulien:
Merged in #239, thanks!

PR #240, opened 2019-08-28
Http evader v14

catenacyber:
Modifies #237

Fixes False Negative outside evader tests with HEAD method (where there is never a response body compared to server code 204 and such where there should not be a response body but there can be)


catenacyber:
Cherry-picked each commit

PR #243, opened 2019-09-12
coverity: makes sure PRId64 is defined

catenacyber:
Fixes new bugs found by coverity

This makes sure stdint.h is included so that `PRId64` and such are defined

Otherwise, it is interpreted as an empty string, and the format string does not use its arguments


catenacyber:
Already cherry-picked

PR #252, opened 2019-09-16
Lzmafix v3

catenacyber:
Fixes our use of LZMA SDK with custom dictionary expanding size

And fixes later processing of data when multiple decompression layers failed

Avoids memory leak from realloc failure

Modifies #251 : 32 bit compatible `size_t` for -1


victorjulien:
I've merged this with a small modification: instead of -1 use SIZE_MAX.

PR #254, opened 2019-09-19
Adds a compression bomb limit

catenacyber:
Triggers when a fixed compression ratio is crossed
And when a configurable output size is reached
Stops decompression and logs an error in this case

Found by oss-fuzz
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=17485

Suricata PR https://github.com/OISF/suricata/pull/4222

Replaces #253 removing a compiler warning about signedness

### DIFF
--- a/htp/htp_config_private.h
+++ b/htp/htp_config_private.h
@@ -345,6 +345,9 @@ struct htp_cfg_t {
 
     /** max memory use by a the lzma decompressor. */
     size_t lzma_memlimit;
+
+    /** max output size for a compression bomb. */
+    int64_t compression_bomb_limit;
 };
 
 #ifdef	__cplusplus
--- a/htp/htp_private.h
+++ b/htp/htp_private.h
@@ -79,6 +79,9 @@ extern "C" {
 
 // 1048576 is 1 Mbyte
 #define HTP_LZMA_MEMLIMIT                   1048576
+//deflate max ratio is about 1000
+#define HTP_COMPRESSION_BOMB_RATIO          4096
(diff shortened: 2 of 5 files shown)


htp/htp_config_private.h
```diff
@@ -345,6 +345,9 @@ struct htp_cfg_t {
 
     /** max memory use by a the lzma decompressor. */
     size_t lzma_memlimit;
+
+    /** max output size for a compression bomb. */
+    int64_t compression_bomb_limit;
```
victorjulien, line 350: I think this size is too big. We won't ever set it to the max. I know size_t also is 64bit (on 64bit archs) and I am generally not happy with how much it is used in the htp API. We should probably make this a uint32_t.

htp/htp_private.h
```diff
@@ -79,6 +79,9 @@ extern "C" {
 
 // 1048576 is 1 Mbyte
 #define HTP_LZMA_MEMLIMIT                   1048576
+//deflate max ratio is about 1000
+#define HTP_COMPRESSION_BOMB_RATIO          4096
```
victorjulien, line 83: Shall we lower this to 2k?

victorjulien (changes requested):

htp/htp_private.h
catenacyber: ok

htp/htp_config_private.h
catenacyber: OK, I use `int32_t` so as not to have signedness warning and I check for overflow in `htp_config_set_compression_bomb_limit`

PR #257, opened 2019-09-19
Adds a compression bomb limit

catenacyber:
Triggers when a fixed compression ratio is crossed
And when a configurable output size is reached
Stops decompression and logs an error in this case

Found by oss-fuzz
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=17485

Suricata PR https://github.com/OISF/suricata/pull/4222

Replaces #254 with code review :
- lower bomb compression ratio to 2048
- use `int32_t` for `compression_bomb_limit` instead of `int64_t` and check for overflow in `htp_config_set_compression_bomb_limit`


catenacyber:
Already cherry-picked with fixed string format

PR #259, opened 2019-09-27
Fuzzit integration v1

catenacyber:
This PR integrates libhtp with fuzzit, a continuous fuzzing platform (such as oss-fuzz)

Travis builds sends built fuzzers to Fuzzit platform for running.

In order to do that, there were some modifications needed to the auto tool compilation scheme, in order to build the fuzzers

PR #260, opened 2019-09-27
fuzz: fuzzit integration

catenacyber:
This PR integrates libhtp with fuzzit, a continuous fuzzing platform (such as oss-fuzz)

Travis builds sends built fuzzers to Fuzzit platform for running.

In order to do that, there were some modifications needed to the auto tool compilation scheme, in order to build the fuzzers

There will be one thing to do for merging that I cannot do :
Use travis secret management https://docs.travis-ci.com/user/encryption-keys/ to secure the API key

cc @yevgenypats

Modifies #259 using local-regression instead of regression for Fuzzit job types in order to have inline feedback into Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,9 @@ compiler:
   - gcc
   - clang
 # Change this to your needs
-script: sh autogen.sh && ./configure && make && make check && make distcheck
+script:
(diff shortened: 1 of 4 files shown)


catenacyber:
Fuzzit is now part of Gitlab

PR #261, opened 2019-10-01
http: support post data decompression

catenacyber:

- Moves some code up in htp_transaction.c
- Creates `htp_tx_req_process_body_data_decompressor_callback` similar to `htp_tx_res_process_body_data_decompressor_callback` but using fields such as `request_entity_len` instead of `response_entity_len`
- Creates common function `htp_tx_init_decompressor`

Remarks :
- Some configuration fields are used for both request and response and not renamed : `response_decompression_layer_limit` and `response_decompression_enabled`
- I am not sure about the behavior of simultaneous body decompression for both a request and a response in the same connection (since we use only one decompressor structure) 
- I made a different use of `response_decompression_enabled` so as not to add a field such as `request_content_encoding_processing` to `htp_tx_t` structure

### DIFF
--- a/htp/htp_transaction.c
+++ b/htp/htp_transaction.c
@@ -238,6 +238,228 @@ void htp_tx_set_user_data(htp_tx_t *tx, void *user_data) {
     tx->user_data = user_data;
 }
 
+/**
+ *  @internal
+ *  @brief split input into tokens separated by "seps"
+ *  @param seps nul-terminated string: each character is a separator
+ */
+static int get_token(const unsigned char *in, size_t in_len, const char *seps,
+    unsigned char **ret_tok_ptr, size_t *ret_tok_len)
+{
+    #if HTP_DEBUG
+    fprintf(stderr, "INPUT %"PRIuMAX, (uintmax_t)in_len);
+    fprint_raw_data(stderr, __func__, in, in_len);
+    #endif
+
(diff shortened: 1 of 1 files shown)


htp/htp_transaction.c
```diff
@@ -574,22 +823,55 @@ htp_status_t htp_tx_req_process_body_data_ex(htp_tx_t *tx, const void *data, siz
...
+            // Send data buffer to the decompressor.
+            tx->connp->out_decompressor->decompress(tx->connp->out_decompressor, &d);
 
+            if (data == NULL) {
+                // Shut down the decompressor, if we used one.
+                htp_connp_destroy_decompressors(tx->connp);
+            }
+            break;
+
+        case HTP_COMPRESSION_NONE:
```
victorjulien, line 856: think this should be the first entry as it is most likely to be true

victorjulien (changes requested):
I would like to see the commit split up so that the code moves are separate. It's too hard to track the changes other than the code moves and function renames.

Request and response should both be able to be compressed. Not sure if this could be complicated by request pipelining. Might be good to create as many pcaps for possible corner cases as possible.

htp/htp_transaction.c
catenacyber: ok

PR #264, opened 2019-10-04

catenacyber:
Fixes #2881

It is less likely that spaces are added in protocol than in URIs

Modifies #264 : rebase with fuzzit integration


victorjulien:
Status: we fear this breaks expectations of users of Suricata and may lead to false negatives.

catenacyber:
Even if this is optional and disabled by default ?

PR #268, opened 2019-10-14
request: finalize with previously buffered stream

catenacyber:
This is rather a draft for discussion than something ready.

The point of this PR is for libhtp to handle requests ending and beginning in the middle of network provided chunks.
That is the added test case 97-requests-cut.t

cf https://github.com/OISF/libhtp/blob/0.5.x/test/test.c#L357 versushttps://github.com/OISF/libhtp/blob/0.5.x/test/test.c#L417

### DIFF
--- a/htp/htp_request.c
+++ b/htp/htp_request.c
@@ -828,40 +828,55 @@ htp_status_t htp_connp_REQ_LINE(htp_connp_t *connp) {
 }
 
 htp_status_t htp_connp_REQ_FINALIZE(htp_connp_t *connp) {
-    size_t bytes_left = connp->in_current_len - connp->in_current_read_offset;
-
-    if (bytes_left > 0) {
-        // If we have more bytes
-        // Either it is request pipelining
-        // Or we interpret it as body data
-        int64_t pos = connp->in_current_read_offset;
-        int64_t mstart = 0;
-        // skip past leading whitespace. IIS allows this
-        while ((pos < connp->in_current_len) && htp_is_space(connp->in_current_data[pos]))
-            pos++;
-        if (pos < connp->in_current_len) {
-            mstart = pos;
-            // The request method starts at the beginning of the
-            // line and ends with the first whitespace character.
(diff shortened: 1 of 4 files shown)


htp/htp_request.c
```diff
@@ -828,40 +828,55 @@ htp_status_t htp_connp_REQ_LINE(htp_connp_t *connp) {
...
 htp_status_t htp_connp_REQ_FINALIZE(htp_connp_t *connp) {
-    size_t bytes_left = connp->in_current_len - connp->in_current_read_offset;
+    if (connp->in_status == HTP_STREAM_CLOSED) {
+        return htp_tx_state_request_complete(connp->in_tx);
+    }
+    for (;;) {//;i < max_read; i++) {
+        IN_PEEK_NEXT(connp);
+        // Have we reached the end of the line? For some reason
+        // we can't test after IN_COPY_BYTE_OR_RETURN */
+        if (connp->in_next_byte == LF || connp->in_next_byte == 0x00)
```
victorjulien, line 838: why test for 0x00?

victorjulien:
Can you update the commit to contain proper commit message:

1. explain the sequence of events leading to the problem
2. explain updated logic

victorjulien:
Can you also add more test cases?

htp/htp_request.c
catenacyber: This was pasted from `htp_connp_REQ_CONNECT_PROBE_DATA`
Not sure why it is there.
It is not in `htp_connp_REQ_LINE`

So, what should I do ? remove them from both functions ?

catenacyber:
> Can you also add more test cases?

Which ones do you expect ?
Do you mean the same double request
```
>>>
GET /?p=%20 HTTP/1.0
User-Agent: Mozilla

GET /?p=%21 HTTP/1.0
User-Agent: Mozilla
```

But cut in different pieces, like
```
>>>
GET /?p=%20 HTTP/1.0
User-Agent: Mozilla

GET /?p
>>>
=%21 HTTP/1.0
User-Agent: Mozilla
```

PR #270, opened 2019-10-24
request: finalize with previously buffered stream

catenacyber:
This is rather a draft for discussion than something ready.

The point of this PR is for libhtp to handle requests ending and beginning in the middle of network provided chunks.
That is the added test case 97-requests-cut.t

Follows #268 by enhancing commit message and removing test for 0x00 as pointed out in the review

### DIFF
--- a/test/files/97-requests-cut.t
+++ b/test/files/97-requests-cut.t
@@ -1,0 +1,9 @@
+>>>
+GET /?p=%20 HTTP/1.0
+User-Agent: Mozilla
+
+G
+>>>
+ET /?p=%21 HTTP/1.0
+User-Agent: Mozilla
+
--- a/htp/htp_request.c
+++ b/htp/htp_request.c
@@ -828,40 +828,55 @@ htp_status_t htp_connp_REQ_LINE(htp_connp_t *connp) {
 }
 
 htp_status_t htp_connp_REQ_FINALIZE(htp_connp_t *connp) {
-    size_t bytes_left = connp->in_current_len - connp->in_current_read_offset;
-
-    if (bytes_left > 0) {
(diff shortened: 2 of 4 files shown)


catenacyber:
After looking at https://redmine.openinfosecfoundation.org/issues/2655 
I am not sure that we should have "fixed" it and pushed the pcap in suricata-verify tests

I am bothered by the fact that the pcap is built in a scenario where the attacker controls both the client and the server, and uses a slightly different protocol than HTTP.
We should rather build a pcap file with an evil client, but a valid HTTP server.
If no valid HTTP server accept these contents from the evil client, we should revert.

test/files/97-requests-cut.t
```diff
@@ -0,0 +1,9 @@
+>>>
+GET /?p=%20 HTTP/1.0
+User-Agent: Mozilla
+
+G
+>>>
+ET /?p=%21 HTTP/1.0
```
victorjulien, line 7: If I remember correctly request pipelining is only supported by HTTP/1.1.

catenacyber: You are right
Do you want me to update the test case ?

victorjulien: Well I guess I'm more surprised about this actually working. Are we simply supporting pipelining for 1.0 as well? This is not RFC compliant as far as I know, but maybe it is seen in the wild?

catenacyber: Yes simply supporting pipelining no matter the protocol version

victorjulien:
When testing this with Suricata master, ``SigTest06`` fails (and other tests too, but those may be a side effect of this first failure).

PR #271, opened 2019-11-06
deadcode: removes it to clean uri decoding

catenacyber:
Removes `htp_uriencoding_normalize_inplace` and dependencies to have only one implementation of url decoding


victorjulien:
Merged in #293, thanks!

PR #269, opened 2019-10-15
test: script to turn htp files into pcaps

catenacyber:
This adds a utility tool to turn htp test files into pcap files so that we can run them with Suricata.

Modifies #267 with even better handling of end of lines delimiters

PR #272, opened 2019-11-15
Request finalize v3

catenacyber:
This is not ready yet and needs more testing.

Follows #270 by fixing same bug for responses and using HTTP/1.1 for real test case with pipelining

### DIFF
--- a/htp/htp_request.c
+++ b/htp/htp_request.c
@@ -828,40 +828,55 @@ htp_status_t htp_connp_REQ_LINE(htp_connp_t *connp) {
 }
 
 htp_status_t htp_connp_REQ_FINALIZE(htp_connp_t *connp) {
-    size_t bytes_left = connp->in_current_len - connp->in_current_read_offset;
-
-    if (bytes_left > 0) {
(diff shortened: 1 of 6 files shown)


victorjulien:
Does this address https://github.com/OISF/libhtp/pull/270#issuecomment-554326682 as well?

victorjulien:
> * LongResponseHeader as the test framework directly returns on error before closing the connection, we do not make it to `tx->request_progress == HTP_REQUEST_COMPLETE`. Should we rather change the test framework ?
>   cf https://github.com/OISF/libhtp/blob/0.5.x/test/test.c#L357 versus https://github.com/OISF/libhtp/blob/0.5.x/test/test.c#L417

Not sure. Can you create a pcap based test (SV test) to see how this affects logging/matching in Suricata?

catenacyber:
> Does this address #270 (comment) as well?

No but #273 does

>  LongResponseHeader : Can you create a pcap based test (SV test) to see how this affects logging/matching in Suricata?

I will with #269

PR #273, opened 2019-11-21

catenacyber:
This is now ready

Follows #273 by having a consistent and seemingly correct handling of responses and requests.
Makes suricata tests pass again

Still a suricata-verify test to craft out of LongResponseHeader

### DIFF
--- a/htp/htp_request.c
+++ b/htp/htp_request.c
@@ -828,39 +828,65 @@ htp_status_t htp_connp_REQ_LINE(htp_connp_t *connp) {
 }
 
 htp_status_t htp_connp_REQ_FINALIZE(htp_connp_t *connp) {
-    size_t bytes_left = connp->in_current_len - connp->in_current_read_offset;
-
-    if (bytes_left > 0) {
-        // If we have more bytes
-        // Either it is request pipelining
-        // Or we interpret it as body data
-        int64_t pos = connp->in_current_read_offset;
-        int64_t mstart = 0;
-        // skip past leading whitespace. IIS allows this
-        while ((pos < connp->in_current_len) && htp_is_space(connp->in_current_data[pos]))
-            pos++;
-        if (pos < connp->in_current_len) {
-            mstart = pos;
-            // The request method starts at the beginning of the
-            // line and ends with the first whitespace character.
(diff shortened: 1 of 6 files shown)


htp/htp_request.c
```diff
@@ -828,40 +828,66 @@ htp_status_t htp_connp_REQ_LINE(htp_connp_t *connp) {
...
-                // Interpret remaining bytes as body data
-                htp_log(connp, HTP_LOG_MARK, HTP_LOG_WARNING, 0, "Unexpected request body");
-                connp->in_tx->request_progress = HTP_REQUEST_BODY;
-                connp->in_state = htp_connp_REQ_BODY_IDENTITY;
-                connp->in_body_data_left = bytes_left;
-                return HTP_OK;
+    if (connp->in_status != HTP_STREAM_CLOSED) {
+        IN_PEEK_NEXT(connp);
+        if (connp->in_next_byte == -1) {
+            return htp_tx_state_request_complete(connp->in_tx);
```
catenacyber, line 834: This may be hacky.
The point is to directly consider the request finalized if we arrive here with no more data to handle
An attacker may trick libhtp to use this path when we do not want to...
This is needed by suricata tests

catenacyber:
So, I just checked with suricata-verify about LongResponseHeader
And here is the diff for eve.json (with and without this PR)
```
384,385c384,385
<       "flows_checked": 1,
<       "flows_notimeout": 1,
---
>       "flows_checked": 0,
>       "flows_notimeout": 0,
390c390
<       "rows_skipped": 65535,
---
>       "rows_skipped": 65536,
393c393
<       "rows_maxlen": 1
---
>       "rows_maxlen": 0
```

How does that look ?

victorjulien:
This fails Suricata-Verify with todays master:
```
===> http-request-invalid: Sub test #1: FAIL : expected 2 matches; got 0 for filter {'count': 2, 'match': {'alert.signature_id': 1, 'event_type': 'alert'}}
```

catenacyber:
> This fails Suricata-Verify with todays master:

Indeed, I have carelessly deleted the htp log which gets interpreted by Suricata as an app layer event.
I am re adding it

PR #274, opened 2019-11-26
time: Adds time limit for pure decompression

catenacyber:
To avoid DOS by small repeated zip bombs

This PR is a first draft.
This already fixes attack found by oss-fuzz https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=18323
Here are the open questions :
- what should be the scope of the time measurement ? I chose the call of `tx->connp->out_decompressor->decompress`
- what is a good default value ? A second ?
- do we have to deal with the race condition of computer time change during our run ?
- Is `htp_decompressor_t` the right structure to add the field `time_spent` ?

### DIFF
--- a/htp/htp_transaction.c
+++ b/htp/htp_transaction.c
@@ -834,8 +834,19 @@ htp_status_t htp_tx_res_process_body_data_ex(htp_tx_t *tx, const void *data, siz
             if (tx->connp->out_decompressor == NULL || tx->connp->out_decompressor->decompress == NULL)
                 return HTP_ERROR;
 
+            struct timeval before, after;
+            gettimeofday(&before, NULL);
             // Send data buffer to the decompressor.
             tx->connp->out_decompressor->decompress(tx->connp->out_decompressor, &d);
+            gettimeofday(&after, NULL);
+            //check for overflows ?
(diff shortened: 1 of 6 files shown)


htp/htp_transaction.c
```diff
@@ -834,8 +834,19 @@ htp_status_t htp_tx_res_process_body_data_ex(htp_tx_t *tx, const void *data, siz
             if (tx->connp->out_decompressor == NULL || tx->connp->out_decompressor->decompress == NULL)
                 return HTP_ERROR;
 
+            struct timeval before, after;
+            gettimeofday(&before, NULL);
             // Send data buffer to the decompressor.
             tx->connp->out_decompressor->decompress(tx->connp->out_decompressor, &d);
+            gettimeofday(&after, NULL);
+            //check for overflows ?
```
victorjulien, line 842: I think you should be able to use `timercmp` here. See `man timercmp` or the existing examples in our code.

catenacyber: Why is that ?
I could rather use `timersub` but I do not think it is better resilient against the overflow from the race condition (if you change the system time during this processing)

victorjulien: Wasn't thinking about time changes, just about a cleaner way to deal with the `struct timeval` then the manual check.

PR #275, opened 2019-11-28
response: missing request - make sure tx is used via out_tx only

anthonylineham:
This prevents a crash which has been seen on several occasions but not reproduced.

### DIFF
--- a/htp/htp_response.c
+++ b/htp/htp_response.c
@@ -1116,6 +1116,10 @@ htp_status_t htp_connp_RES_IDLE(htp_connp_t *connp) {
         if (connp->out_tx == NULL) {
             return HTP_ERROR;
         }
+        // Creating a tx instance automatically assigns it to in_tx. Clear it to avoid it
+        // being used out of context
+        connp->in_tx = NULL;
+
         connp->out_tx->parsed_uri = htp_uri_alloc();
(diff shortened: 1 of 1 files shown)


catenacyber:
Thanks for this contribution.
This looks indeed a problem.

Why did you choose this patch over modifying `htp_connp_tx_create` by removing 
`connp->in_tx = tx;` from it ?

anthonylineham:
Thanks for suggesting this. I was going for a change that was minimally disruptive. But I can see that your suggestion would be safe as the use-case in htp_request.c gets assigned to connp->in_tx directly.
I'm new to github. Should I apply a new patch on top of my current one or something else?
Thanks

catenacyber:
> Should I apply a new patch on top of my current one or something else?

The custom for this project is to open a new Pull Request, where you quote this one with `#275`, explain what you modified, and close this Pull Request.
You could use a git branch named response-in_tx-v2

catenacyber:
See https://github.com/OISF/libhtp/pull/278 for instance

catenacyber:
@anthonylineham should we close this PR as discussion continus in #279 ?

anthonylineham:
@catenacyber yes, I'll do that now.

PR #276, opened 2019-11-28
htp: Prevent memory leak with repeated content-type headers

anthonylineham:
Before saving the content-type value free any existing string.
This can happen in cases where only one side of the HTTP session
is seen and there are multiple requests in the one session.

### DIFF
--- a/htp/htp_response.c
+++ b/htp/htp_response.c
@@ -634,6 +634,10 @@ htp_status_t htp_connp_RES_BODY_DETERMINE(htp_connp_t *connp) {
         // We have a response body
         htp_header_t *ct = htp_table_get_c(connp->out_tx->response_headers, "content-type");
         if (ct != NULL) {
+            if (connp->out_tx->response_content_type) {
+                bstr_free(connp->out_tx->response_content_type);
+            }
+
             connp->out_tx->response_content_type = bstr_dup_lower(ct->value);
             if (connp->out_tx->response_content_type == NULL) return HTP_ERROR;
 


catenacyber:
Thank you @anthonylineham for this contribution.
Could you provide a test case to reproduce the bug ? That could be a pcap, a file in .t format (see the ones in the test/files directory), or whatever you prefer :-)

> This can happen in cases where only one side of the HTTP session is seen 

Which one side ? The client to server or server to client ?
It looks like the server to client as you patch htp_response.c but you mention multiple requests

> there are multiple requests in the one session.

Then, there should be multiple transactions, each of them having only one `response_content_type`. Is it the case ?

anthonylineham:
Sorry for the brief description. We're using Suricata in a configuration with NAT. Hence Suricata sees the client->server and server->client sides of the flow independently (we're using async-oneside mode).
It is the server->client side that causes the problem - as you guessed.
Because Suricata is seeing each side independently, the flow handling the server->client side doesn't see the requests at all. I think this is reason there are not multiple transactions. 

I will work on getting a pcap that will reproduce this.

catenacyber:
Thanks, we will be waiting for the pcap. The bug may be deeper and need more patches elsewhere

anthonylineham:
In the process of getting the pcap ready I've found the that issue has been fixed.
I was debugging with 0.5.25 plus a cherry-pick of https://github.com/OISF/libhtp/commit/0c77ea62a7b3d12a141e29b6d949b839e073f597
This exhibited the problem and I could reproduce it readily.
However, when I tried it with Victor's follow-up patch https://github.com/OISF/libhtp/commit/073a50816d73df0e7c49883b13ba7ac380f394ee the problem didn't occur. I've narrowed it down to the changes in htp_connp_RES_IDLE():

>          connp->in_state = htp_connp_REQ_FINALIZE;
>          ...
>         // We've used one transaction
>         connp->out_next_tx_index++;
This all makes sense given the explanation of what out_next_tx_index does.

On one hand I think my change is not needed now. On the other it seems like it wouldn't hurt to have it there just in case. What do you think?

catenacyber:
> What do you think?

I am glad that the bug was already fixed.
If it is not needed, I think we do not want to waste 1 CPU cycle checking for this.
Maybe we can turn this into an `assert` so that it is not used in production but can be tested anyways

anthonylineham:
Closing this pull request as the issue is already resolved in currently 0.5.x

PR #278, opened 2019-11-28

catenacyber:
This is now ready

Follows #277 by doing the log
ie `htp_log(connp, HTP_LOG_MARK, HTP_LOG_WARNING, 0, "Unexpected request body");` in `htp_connp_REQ_FINALIZE` only if there are more than whitespaces in the unexpected body


catenacyber:
All suricata-verify tests pass locally now :-)

PR #279, opened 2019-11-29
response: missing request - make sure tx is used via out_tx only

anthonylineham:
The return value from htp_connp_tx_create() already gets assigned to
the right place. It's not appropriate to automatically assign it
to in_tx as it may have been created for a response with out request.

This prevents a crash which has been seen but not reproduced.

I previously made a fix (#275) for this issue by clearing the connp->in_tx in htp_connp_RES_IDLE(), but it was pointed out that the automatic assignment in htp_connp_tx_create() was not needed and could be removed instead.

### DIFF
--- a/htp/htp_connection_parser.c
+++ b/htp/htp_connection_parser.c
@@ -194,8 +194,6 @@ htp_tx_t *htp_connp_tx_create(htp_connp_t *connp) {
     htp_tx_t *tx = htp_tx_create(connp);
     if (tx == NULL) return NULL;
 
-    connp->in_tx = tx;   
-
     htp_connp_in_reset(connp);
 
     return tx;
--- a/test/test_hybrid.cpp
+++ b/test/test_hybrid.cpp
@@ -242,7 +242,7 @@ protected:
  */
 TEST_F(HybridParsing, GetTest) {
     // Create a new LibHTP transaction
-    htp_tx_t *tx = htp_connp_tx_create(connp);
+    htp_tx_t *tx = connp->in_tx =htp_connp_tx_create(connp);
     ASSERT_TRUE(tx != NULL);
 
     // Configure user data and callbacks
@@ -394,7 +394,7 @@ TEST_F(HybridParsing, GetTest) {
  */
 TEST_F(HybridParsing, PostUrlecodedTest) {
     // Create a new LibHTP transaction
-    htp_tx_t *tx = htp_connp_tx_create(connp);
+    htp_tx_t *tx = connp->in_tx = htp_connp_tx_create(connp);
     ASSERT_TRUE(tx != NULL);
 
     // Request begins
@@ -504,7 +504,7 @@ static void HybridParsing_CompressedResponse_Setup(htp_tx_t *tx) {
  */
 TEST_F(HybridParsing, CompressedResponse) {
     // Create a new LibHTP transaction
-    htp_tx_t *tx = htp_connp_tx_create(connp);
+    htp_tx_t *tx = connp->in_tx = htp_connp_tx_create(connp);
     ASSERT_TRUE(tx != NULL);
 
     HybridParsing_CompressedResponse_Setup(tx);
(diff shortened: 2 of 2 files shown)


anthonylineham:
@catenacyber Do you know why this build failed? 
All I can find is
 ../test-driver: line 107:  8578 Segmentation fault      (core dumped) "$@" > $log_file 2>&1
but I can't work out what this relates to. Any ideas

catenacyber:
> Do you know why this build failed?

Did you try to run `make check` locally ?
It seems that one of the tests segfaults.
You can then try to read test/test_suite.log to find which one

anthonylineham:
Thanks for the suggestion. I ran the test locally and found the problem. The tests, unlike the actual library, rely on connp->in_tx being set in htp_connp_tx_create().

I can change the affected tests to assign the value returned from htp_connp_tx_create() to connp->in_tx. Do think that will be acceptable? Or should I go back to my original fix?

catenacyber:
> The tests, unlike the actual library, rely on connp->in_tx being set in htp_connp_tx_create().

Are these specific tests or the test framework ?
I guess that we need to change the tests.
Do you want to investigate further or let me do it ?

anthonylineham:
It's specific tests. They are all in test_hybrid.cpp. 
Thanks for the offer of investigating! I've got it sorted, I think. The fix is identical in each case. I will push the fix to this pull request soon.

test/test_hybrid.cpp
```diff
@@ -242,7 +242,7 @@ class HybridParsing : public testing::Test {
  */
 TEST_F(HybridParsing, GetTest) {
     // Create a new LibHTP transaction
-    htp_tx_t *tx = htp_connp_tx_create(connp);
+    htp_tx_t *tx = connp->in_tx =htp_connp_tx_create(connp);
```
catenacyber: nit : missing space

catenacyber (approved):
This looks good to me.

If you can reproduce the crash, we would be happy to turn it into a suricata-verify test

htp/htp_connection_parser.c
```diff
@@ -194,8 +194,6 @@ htp_tx_t *htp_connp_tx_create(htp_connp_t *connp) {
     htp_tx_t *tx = htp_tx_create(connp);
     if (tx == NULL) return NULL;
 
-    connp->in_tx = tx;   
-
     htp_connp_in_reset(connp);
```
catenacyber, line 197: Do we still need this line ? (now that we do not assign `connp->in_tx`)
Should it be moved to the place in the code where `connp->in_tx` is assigned ?

catenacyber (changes requested):
Well, there seems to be cases where this feature is used cf https://github.com/OISF/libhtp/blob/0.5.x/htp/htp_transaction.c#L1079

Besides, we need `make check` to work on Suricata with these changes (and they seem to be making some NULL dereference, or use of a freed pointer)

So, this needs further investigation.
Can you reproduce the bug and share a pcap ?

htp/htp_connection_parser.c
anthonylineham: Just so I understand correctly, are you talking about the call to htp_connp_in_reset()?

test/test_hybrid.cpp
anthonylineham: Whoops! I've pushed another patch for this

anthonylineham:
I'll have to review the core dump info again and see what I can work out.

htp/htp_connection_parser.c
catenacyber: Indeed

test/test_hybrid.cpp
catenacyber: thanks

catenacyber:
You can take a look at https://github.com/OISF/libhtp/pull/283/files#diff-a29b27846643a6568e8a05fde36e612cR1143 which may be a related fix

anthonylineham:
I'm going to abandon this as I don't have a reproduction method so I don't think I can progress this.

catenacyber:
Thanks, let us know if you reproduce it sometime

PR #280, opened 2019-12-02
cmake: compile scheme available

catenacyber:
This is a draft PR, not to be merged right now.

This adds a way to compile libhtp with cmake.
This is very basic (lacks checks for packages and such) but it works on a good system.


catenacyber:
Closing as this does not seem a priority for now

PR #283, opened 2019-12-05

catenacyber:
This is now ready

Follows #278 by adding another way to complete the request part of a transaction : when we see a second response (before reassigning to a new `in_tx`)

### DIFF
--- a/htp/htp_response.c
+++ b/htp/htp_response.c
@@ -1070,21 +1070,49 @@ size_t htp_connp_res_data_consumed(htp_connp_t *connp) {
 }
 
 htp_status_t htp_connp_RES_FINALIZE(htp_connp_t *connp) {
-    int bytes_left = connp->out_current_len - connp->out_current_read_offset;
-    unsigned char * data = connp->out_current_data + connp->out_current_read_offset;
+    if (connp->out_status != HTP_STREAM_CLOSED) {
+        OUT_PEEK_NEXT(connp);
+        if (connp->out_next_byte == -1) {
+            return htp_tx_state_response_complete_ex(connp->out_tx, 0);
+        }
+        if (connp->out_next_byte != LF || connp->out_current_consume_offset >= connp->out_current_read_offset) {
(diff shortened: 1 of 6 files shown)


htp/htp_response.c
```diff
@@ -1070,21 +1070,49 @@ size_t htp_connp_res_data_consumed(htp_connp_t *connp) {
 }
 
 htp_status_t htp_connp_RES_FINALIZE(htp_connp_t *connp) {
-    int bytes_left = connp->out_current_len - connp->out_current_read_offset;
-    unsigned char * data = connp->out_current_data + connp->out_current_read_offset;
+    if (connp->out_status != HTP_STREAM_CLOSED) {
+        OUT_PEEK_NEXT(connp);
+        if (connp->out_next_byte == -1) {
+            return htp_tx_state_response_complete_ex(connp->out_tx, 0);
```
catenacyber, line 1076: This does not seem to be always the right thing to do, but is needed by unit tests

victorjulien: When would this not be correct and why?

victorjulien:
If we look for an entire request line, does it matter how long that request line is? Also, does it matter if the request line itself might not be complete yet in the data that is fed to us?

htp/htp_response.c
catenacyber: When the end of a packet is `LF` but the next packet is not the start of a new request for instance.

For instance, in the private example you shared with me, is packet 7 a new response or continuation body of the previous response ?

catenacyber: And this may not be correct because we want to interpret this junk as body, not as a new request/reply

catenacyber:
> If we look for an entire request line, does it matter how long that request line is? 

Not really except for memory consumption (but it is already the case) : we buffer until a new line arrives or we hit the buffering limit `field_limit_hard`

> Also, does it matter if the request line itself might not be complete yet in the data that is fed to us?

In this code no, we only check the method, but we could add furthers checks (especially the protocol part at the end of the request line)

PR #284, opened 2019-12-08
time: Adds time limit for pure decompression

catenacyber:
To avoid DOS by small repeated zip bombs

This PR is a first draft.
It is another version of #274 dealing with https://github.com/OISF/libhtp/pull/274#discussion_r355128131

> Wasn't thinking about time changes, just about a cleaner way to deal with the `struct timeval` then the manual check.

I am not sure this change improves the code, but I may not have fully understood the comment
So I leave #274 open for now

### DIFF
--- a/htp/htp_config.h
+++ b/htp/htp_config.h
@@ -443,6 +443,15 @@ void htp_config_set_lzma_memlimit(htp_cfg_t *cfg, size_t memlimit);
 void htp_config_set_compression_bomb_limit(htp_cfg_t *cfg, size_t bomblimit);
 
 /**
+ * Configures the maximum compression bomb time LibHTP will decompress.
+ *
+ * @param[in] cfg
+ * @param[in] seclimit
+ * @param[in] useclimit
+ */
+void htp_config_set_compression_time_limit(htp_cfg_t *cfg, size_t seclimit, size_t useclimit);
+
+/**
  * Configures the desired log level.
  * 
  * @param[in] cfg
--- a/htp/htp_transaction.c
+++ b/htp/htp_transaction.c
@@ -834,8 +834,21 @@ htp_status_t htp_tx_res_process_body_data_ex(htp_tx_t *tx, const void *data, siz
             if (tx->connp->out_decompressor == NULL || tx->connp->out_decompressor->decompress == NULL)
                 return HTP_ERROR;
 
+            struct timeval before, after, duration;
+            gettimeofday(&before, NULL);
             // Send data buffer to the decompressor.
             tx->connp->out_decompressor->decompress(tx->connp->out_decompressor, &d);
+            gettimeofday(&after, NULL);
+            //check for overflows ?
+            timersub(&after, &before, &duration);
+            timeradd(&tx->connp->out_decompressor->time_spent, &duration, &tx->connp->out_decompressor->time_spent);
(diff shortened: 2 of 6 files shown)


htp/htp_transaction.c
```diff
@@ -834,8 +834,21 @@ htp_status_t htp_tx_res_process_body_data_ex(htp_tx_t *tx, const void *data, siz
             if (tx->connp->out_decompressor == NULL || tx->connp->out_decompressor->decompress == NULL)
                 return HTP_ERROR;
 
+            struct timeval before, after, duration;
+            gettimeofday(&before, NULL);
             // Send data buffer to the decompressor.
             tx->connp->out_decompressor->decompress(tx->connp->out_decompressor, &d);
+            gettimeofday(&after, NULL);
+            //check for overflows ?
```
victorjulien, line 842: Did not compile test this, but what about something like
```c
struct timeval start, limit, after;
memset(&limit, 0, sizeof(limit));
gettimeofday(&start, NULL);
timeradd(&start, &tx->connp->out_decompressor->time_spent, &limit);

// Send data buffer to the decompressor.
tx->connp->out_decompressor->decompress(tx->connp->out_decompressor, &d);
gettimeofday(&after, NULL);

if ( timercmp(&after, &limit, >) ) {
 // spent too much time
}
```

htp/htp_config.h
```diff
@@ -442,6 +442,15 @@ void htp_config_set_lzma_memlimit(htp_cfg_t *cfg, size_t memlimit);
...
 void htp_config_set_compression_bomb_limit(htp_cfg_t *cfg, size_t bomblimit);
 
+/**
+ * Configures the maximum compression bomb time LibHTP will decompress.
+ *
+ * @param[in] cfg
+ * @param[in] seclimit
+ * @param[in] useclimit
+ */
+void htp_config_set_compression_time_limit(htp_cfg_t *cfg, size_t seclimit, size_t useclimit);
```
victorjulien, line 452: there really isn't a scenario I can think of that would have > second level timeouts that make sense, so it seems we could do with a smaller type here and use it to create a local `struct timeval` in the code that uses it

htp/htp_transaction.c
catenacyber: Ah ! I guess I finally understand our difference in thinking :-)

I propose to cumulate the time spent in _multiple_ calls of `decompress` for one connection
While you measure and limit only the time spent in _one_ call to `decompress`

Hence the general question : what should be the scope of the time measurement ? 
I think it safer to cumulate to avoid some evasion by attacker splitting into smaller chunks...
But I am not sure. What is your argument for one call only ?

victorjulien: Actually I missed this about your approach. I was only thinking about a way to not have to worry about the timersub overflow case when time might change.

victorjulien: I do agree we should track the cumulative cost.

htp/htp_config.h
catenacyber: ok, enforced a hard limit of one second

htp/htp_transaction.c
catenacyber: ok : added a sanity check for timersub overflow and kept the cumulative cost

PR #289, opened 2020-01-22
Time limit v4

catenacyber:
To avoid DOS by small repeated zip bombs or big lzma bomb

Modifies #288 with
- adding checks into `htp_tx_res_process_body_data_decompressor_callback` (called  by `decompress`)
- Checks are not done every iteration of callback but every `HTP_COMPRESSION_TIME_FREQ_TEST` times
- `time_before` is moved to structure to be accessible from `htp_tx_res_process_body_data_decompressor_callback`

### DIFF
--- a/htp/htp_transaction.c
+++ b/htp/htp_transaction.c
@@ -834,8 +834,23 @@ htp_status_t htp_tx_res_process_body_data_ex(htp_tx_t *tx, const void *data, siz
             if (tx->connp->out_decompressor == NULL || tx->connp->out_decompressor->decompress == NULL)
                 return HTP_ERROR;
 
+            struct timeval before, after, duration;
+            gettimeofday(&before, NULL);
             // Send data buffer to the decompressor.
             tx->connp->out_decompressor->decompress(tx->connp->out_decompressor, &d);
+            gettimeofday(&after, NULL);
+            // sanity check for race condition if system time changed
+            if ( timercmp(&after, &before, >) ) {
+                timersub(&after, &before, &duration);
+                timeradd(&tx->connp->out_decompressor->time_spent, &duration, &tx->connp->out_decompressor->time_spent);
+                if ( timercmp(&tx->connp->out_decompressor->time_spent, &tx->connp->cfg->compression_time_limit, >) ) {
+                    htp_log(tx->connp, HTP_LOG_MARK, HTP_LOG_ERROR, 0,
+                            "Compression bomb: spent %"PRId64"s and "PRId64" us decompressing",
+                            tx->connp->out_decompressor->time_spent.tv_sec,
+                            tx->connp->out_decompressor->time_spent.tv_usec);
+                    return HTP_ERROR;
+                }
+            }
 
             if (data == NULL) {
                 // Shut down the decompressor, if we used one.
--- a/htp/htp_transaction.c
+++ b/htp/htp_transaction.c
@@ -789,6 +789,25 @@ static htp_status_t htp_tx_res_process_body_data_decompressor_callback(htp_tx_da
     // Invoke all callbacks.
     htp_status_t rc = htp_res_run_hook_body_data(d->tx->connp, d);
     if (rc != HTP_OK) return HTP_ERROR;
+    d->tx->connp->out_decompressor->nb_callbacks++;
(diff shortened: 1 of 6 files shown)


catenacyber:
Fixes new timeout found by oss-fuzz https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=20234

htp/htp_transaction.c
```diff
@@ -789,6 +789,25 @@ static htp_status_t htp_tx_res_process_body_data_decompressor_callback(htp_tx_da
...
     htp_status_t rc = htp_res_run_hook_body_data(d->tx->connp, d);
     if (rc != HTP_OK) return HTP_ERROR;
+    d->tx->connp->out_decompressor->nb_callbacks++;
+    if (d->tx->connp->out_decompressor->nb_callbacks % HTP_COMPRESSION_TIME_FREQ_TEST == 0) {
+        struct timeval after, duration;
+        gettimeofday(&after, NULL);
+        // sanity check for race condition if system time changed
+        if ( timercmp(&after, &d->tx->connp->out_decompressor->time_before, >) ) {
+            timersub(&after, &d->tx->connp->out_decompressor->time_before, &duration);
+            timeradd(&d->tx->connp->out_decompressor->time_spent, &duration, &d->tx->connp->out_decompressor->time_spent);
```
victorjulien, line 799: is there a reason for using `duration` here, instead of just calculating `time_spent` directly from `time_before` and `after`?

catenacyber: Yes, cumulate time spent over multiple calls to `decompress` for one response
This may be over cautious... But I was worried to forget one corner case.
What do you think ?

victorjulien: Maybe I am misunderstanding the overall flow. But I see that `htp_tx_res_process_body_data_ex` we set the `time_before` and then after running the decompress check if it took too long. We do this by storing the diff between before and after in the local variable `duration`, which is then added to `time_spent`.

But then in `htp_tx_res_process_body_data_decompressor_callback`, which I assume is called (indirectly) from `htp_tx_res_process_body_data_ex`, does almost the same, except it does reuse the `time_before` value.

Does this mean `duration` will always contain the full time spent?

Does this mean that `time_spent` is actually too high, as both `htp_tx_res_process_body_data_ex` and `htp_tx_res_process_body_data_decompressor_callback` increase it while using the same `time_before` as their base?

catenacyber: > Does this mean duration will always contain the full time spent?

No, `duration` contains the duration time for one "atomic" interval between two `gettimeofday`

> Does this mean that time_spent is actually too high, as both htp_tx_res_process_body_data_ex and htp_tx_res_process_body_data_decompressor_callback increase it while using the same time_before as their base?

No because `htp_tx_res_process_body_data_decompressor_callback` updates `time_before`
That is the _trick_

htp/htp_transaction.c
```diff
@@ -789,6 +789,25 @@ static htp_status_t htp_tx_res_process_body_data_decompressor_callback(htp_tx_da
     // Invoke all callbacks.
     htp_status_t rc = htp_res_run_hook_body_data(d->tx->connp, d);
     if (rc != HTP_OK) return HTP_ERROR;
+    d->tx->connp->out_decompressor->nb_callbacks++;
+    if (d->tx->connp->out_decompressor->nb_callbacks % HTP_COMPRESSION_TIME_FREQ_TEST == 0) {
+        struct timeval after, duration;
+        gettimeofday(&after, NULL);
+        // sanity check for race condition if system time changed
+        if ( timercmp(&after, &d->tx->connp->out_decompressor->time_before, >) ) {
+            timersub(&after, &d->tx->connp->out_decompressor->time_before, &duration);
```
victorjulien, line 798: Looks like I've put you on the wrong path here. `timeradd` and `timersub` are not available on all of our platforms, It's missing on Windows (MinGW). `timercmp` *is* available already in use in Suricata in a few places.

We could see if we can add in a properly licensed implementation from somewhere, or we can do back to code that doesn't use these calls.

catenacyber: I will go back to first version where we do not use these functions

PR #290, opened 2020-02-09
Fixed empty tx bug in REQ_IDLE

jferrant:
In `htp_connp_REQ_FINALIZE`, we unread the last end of line so that subsequent calls to `htp_connp_REQ_LINE` work. However, whenever we enter `htp_connp_REQ_IDLE` state, the check `IN_TEST_NEXT_BYTE_OR_RETURN(connp)` always passes even when there is no more data beyond the CRLF creating an empty transaction. `htp_connp_REQ_IDLE` now explicitly checks to make sure that if we only have two bytes of data,  those two bytes are not simply the CRLF before starting a new transaction.

### DIFF
--- a/htp/htp_request.c
+++ b/htp/htp_request.c
@@ -929,6 +929,13 @@ htp_status_t htp_connp_REQ_IDLE(htp_connp_t * connp) {
     // connection.
     IN_TEST_NEXT_BYTE_OR_RETURN(connp);
 
+    // If there is only CRLF left, we just unread last end of line so that REQ_LINE works.
+    // Need more data before starting a new request.
+    if (connp->in_current_read_offset + 2 == connp->in_current_len &&
+        connp->in_current_data[connp->in_current_read_offset] == CR &&
+        connp->in_current_data[connp->in_current_read_offset + 1] == LF)
+        return HTP_DATA;
+
     connp->in_tx = htp_connp_tx_create(connp);
     if (connp->in_tx == NULL) return HTP_ERROR;
 


catenacyber:
Thanks for this report.
Can you share a pcap with this behavior ?

jferrant:
> Thanks for this report.
> Can you share a pcap with this behavior ?

Since github says it doesn't handle pcaps, I base64ed a pcap and attached it as a txt file. Let me know if for some reason it doesn't replicate the bug for you. The bug happens with pcaps that have an extra \r\n after the headers. For example:
GET test.txt HTTP/1.1\r\nUser-Agent: BLAH\r\n\r\n\r\n

[htp-extra-req.txt](https://github.com/OISF/libhtp/files/4389567/htp-extra-req.txt)

catenacyber:
Thanks for the pcap

By the way, you can use the script test/pcaptohtp.py (see the comments in the code to see how it works) to transform the pcap into a libhtp test (text file)

I now understand your case.
But I think the fix should rather be in `htp_connp_REQ_FINALIZE`

```
diff --git a/htp/htp_request.c b/htp/htp_request.c
index 3756de3..2a87354 100644
--- a/htp/htp_request.c
+++ b/htp/htp_request.c
@@ -882,7 +882,12 @@ htp_status_t htp_connp_REQ_FINALIZE(htp_connp_t *connp) {
             htp_status_t rc = htp_tx_req_process_body_data_ex(connp->in_tx, data, len);
             htp_connp_req_clear_buffer(connp);
             return rc;
-        }
+        } // else continue
+    } else {
+        //empty whitespace line
+        htp_status_t rc = htp_tx_req_process_body_data_ex(connp->in_tx, data, len);
+        htp_connp_req_clear_buffer(connp);
+        return rc;
     }
     //else
     //unread last end of line so that REQ_LINE works
```

What do you think of this patch ?

jferrant:
Yes, this is a much better fix! No issues on my side :)

victorjulien:
Replaced by #291

PR #294, opened 2020-04-25
htp.pc.in: add -lz to Libs.private

ffontaine:
zlib is a mandatory dependency so add it to Libs.private otherwise
static linking of packages linking with htp (e.g. suricata) will fail.

[Retrieved from:
https://git.buildroot.net/buildroot/tree/package/libhtp/0002-htp.pc.in-add-lz-to-Libs.private.patch]
Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/htp.pc.in
+++ b/htp.pc.in
@@ -7,6 +7,6 @@ Name: @PACKAGE_NAME@
 Description: A security-aware HTTP parser, designed for use in IDS/IPS and WAF products.
 Version: @PACKAGE_VERSION@
 Libs: -L${libdir} -lhtp
-Libs.private: @LIBICONV@
+Libs.private: -lz @LIBICONV@
 Cflags: -I${includedir} -I${libdir}/htp/include
 


catenacyber:
Thanks.
I am not familiar with `.pc.in` files.
Which build system uses this file and how ?

ffontaine:
`.pc.in` files are used to generate `.pc` files which can be used by other packages (such as suricata) to retrieve the library dependencies thanks to **pkg-config**. These files are especially useful to retrieve static dependencies thanks to `Libs.Private`. Indeed, when linking statically an application you must retrieve the complete library list and link with them in the correct order. pkg-config can be called from autotools, cmake or meson buildsystems. You can find more information about pkg-config here: https://people.freedesktop.org/~dbn/pkg-config-guide.html and about static linking here: https://eli.thegreenplace.net/2013/07/09/library-order-in-static-linking

catenacyber:
Thanks @ffontaine 

Looks good to me

catenacyber:
Merged in https://github.com/OISF/libhtp/pull/308

PR #302, opened 2020-09-04
decompression: limit lzma layers to one

catenacyber:
and log an error if there are more

cf https://redmine.openinfosecfoundation.org/issues/3776

Modifies #301 with up to date rebase

### DIFF
--- a/htp/htp_transaction.c
+++ b/htp/htp_transaction.c
@@ -1280,6 +1280,7 @@ htp_status_t htp_tx_state_response_headers(htp_tx_t *tx) {
         } else {
             int layers = 0;
             htp_decompressor_t *comp = NULL;
+            int nblzma = 0;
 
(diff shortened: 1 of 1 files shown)


victorjulien:
I thought we wanted to make this configurable, with a default of 1.

Also, can this check be bypassed by the actual body decompression logic? I think at least for gzip and deflate we just try to decompress the data irrespective of the headers. Not sure about lzma.

catenacyber:
> I thought we wanted to make this configurable, with a default of 1.

Ok, doing this change

> Also, can this check be bypassed by the actual body decompression logic? I think at least for gzip and deflate we just try to decompress the data irrespective of the headers. Not sure about lzma.

This check is not bypassed by current detection logic.
`htp_gzip_decompressor_restart` tries indeed to probe for deflate and gzip, but not lzma

PR #303, opened 2020-09-09
decompression: configurable limit for lzma layers

catenacyber:
Default value is one.
Log an error if there are more lzma layers.

Replaces https://github.com/OISF/libhtp/pull/302 by making the limit configurable

https://redmine.openinfosecfoundation.org/issues/3776
cf https://github.com/OISF/suricata/pull/5365

### DIFF
--- a/htp/htp_config_private.h
+++ b/htp/htp_config_private.h
@@ -343,6 +343,9 @@ struct htp_cfg_t {
     /** How many layers of compression we will decompress (0 => no limit). */
     int response_decompression_layer_limit;
 
+    /** How many layers of compression we will decompress (0 => no lzma). */
+    int response_lzma_layer_limit;
+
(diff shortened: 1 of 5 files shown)


htp/htp_config_private.h
```diff
@@ -343,6 +343,9 @@ struct htp_cfg_t {
     /** How many layers of compression we will decompress (0 => no limit). */
     int response_decompression_layer_limit;
 
+    /** How many layers of compression we will decompress (0 => no lzma). */
+    int response_lzma_layer_limit;
```
victorjulien, line 347: I think this will break the ABI. Please append the new field so the layout in memory doesn't change

catenacyber: Done in next PR

PR #307, opened 2020-09-09
Http expect100 3880 v3

catenacyber:
https://redmine.openinfosecfoundation.org/issues/3880

Handles expect 100-continue header And limits some logging

Replaces #306 with needed rebasing


catenacyber:
Merged in https://github.com/OISF/libhtp/pull/308

PR #310, opened 2020-09-11
request: handle body decompression

catenacyber:

- Creates `htp_tx_req_process_body_data_decompressor_callback` similar to `htp_tx_res_process_body_data_decompressor_callback` but using fields such as `request_entity_len` instead of `response_entity_len` (not sure if we should generalize)
- Modifies `htp_connp_destroy_decompressors` to call `htp_tx_res_destroy_decompressors` and not the other way around

Modifies #262 with rebase

Test in https://github.com/OISF/suricata-verify/pull/329

### DIFF
--- a/htp/htp_config_private.h
+++ b/htp/htp_config_private.h
@@ -175,6 +175,9 @@ struct htp_cfg_t {
     /** Whether to decompress compressed response bodies. */
     int response_decompression_enabled;
 
+    /** Whether to decompress compressed request bodies. */
+    int request_decompression_enabled;
+
     /** Not fully implemented at the moment. */
     char *request_encoding;
 
--- a/htp/htp_connection_parser_private.h
+++ b/htp/htp_connection_parser_private.h
@@ -253,6 +253,9 @@ struct htp_connp_t {
     /** Response decompressor used to decompress response body data. */
     htp_decompressor_t *out_decompressor;
 
+    /** Request decompressor used to decompress request body data. */
+    htp_decompressor_t *req_decompressor;
+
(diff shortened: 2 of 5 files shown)


htp/htp_config_private.h
```diff
@@ -175,6 +175,9 @@ struct htp_cfg_t {
     /** Whether to decompress compressed response bodies. */
     int response_decompression_enabled;
 
+    /** Whether to decompress compressed request bodies. */
+    int request_decompression_enabled;
```
victorjulien, line 179: breaks ABI

htp/htp_connection_parser_private.h
```diff
@@ -253,6 +253,9 @@ struct htp_connp_t {
     /** Response decompressor used to decompress response body data. */
     htp_decompressor_t *out_decompressor;
 
+    /** Request decompressor used to decompress request body data. */
+    htp_decompressor_t *req_decompressor;
```
victorjulien, line 257: breaks ABI

victorjulien:
I think this should be disabled by default for Suricata 5 and 6. Then we'll enable it by default from inside suricata 7.

htp/htp_config_private.h
catenacyber: ok

htp/htp_connection_parser_private.h
catenacyber: ok

catenacyber:
Replaced by https://github.com/OISF/libhtp/pull/318

PR #315, opened 2021-01-15
Gap finalize before 4219 v2

catenacyber:
Completing fix for https://redmine.openinfosecfoundation.org/issues/4219

- fixing the fuzz target for gaps
- handling the in data receiver hook as the out one
- fixing covert unchecked return values (present before)


catenacyber:
Replaced by https://github.com/OISF/libhtp/pull/316

PR #318, opened 2021-01-21
request: handle body decompression

catenacyber:

- Creates `htp_tx_req_process_body_data_decompressor_callback` similar to `htp_tx_res_process_body_data_decompressor_callback` but using fields such as `request_entity_len` instead of `response_entity_len` (not sure if we should generalize)
- Modifies `htp_connp_destroy_decompressors` to call `htp_tx_res_destroy_decompressors` and not the other way around

Modifies #310 with disabling by default (for previous suricata versions) and not breaking ABI

Test in https://github.com/OISF/suricata-verify/pull/329


victorjulien:
Merged in #319, thanks!

PR #320, opened 2021-01-26
coverity: fix format string for int32

catenacyber:
time_spent is defined as `int32_t` in htp_decompressors.h


catenacyber:
ping @victorjulien

PR #321, opened 2021-02-05
fix invalid unsigned arithmetic.

ihsinme:
I believe your checks are not correct.
in my opinion they are equivalent to checks !=.
I suggest a simple fix.

### DIFF
--- a/htp/htp_urlencoded.c
+++ b/htp/htp_urlencoded.c
@@ -65,7 +65,7 @@ static void htp_urlenp_add_field_piece(htp_urlenp_t *urlenp, const unsigned char
             // The current field consists of more than once piece, we have to use the string builder.
 
             // Add current piece to string builder.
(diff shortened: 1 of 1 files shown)


catenacyber:
Thanks.
Do you have a test case to check that behavior is correct with this patch ?

ihsinme:
no, I have no test to check.
I thought this was a very simple and straightforward fix.

if a test is needed give me an example.

catenacyber:
The fix looks reasonable indeed.

But I would like to know how libhtp can now fail, and add a test to check that this does not happen again.

> if a test is needed give me an example.

You can look in the test directory.
You can for instance create a sample test file in test/files and use it in test/test_main.cpp

ihsinme:
good day.
I'm sorry to be late with the reply.
After re-examining your code, I realized that your comparison will also work correctly.
since `endpos` can never be less than `startpos`.

PR #322, opened 2021-02-17
time: do not put the whole stream in error state

catenacyber:
in case we reached some time limit
just set the decompressor in passthrough mode

https://redmine.openinfosecfoundation.org/issues/4331


victorjulien:
Merged in #323, thanks!

PR #326, opened 2021-04-07
chunk: consume empty lines when parsing chunk length

catenacyber:
so as to avoid quadratic complexity

https://redmine.openinfosecfoundation.org/issues/4401


victorjulien:
Merged in #327

PR #331, opened 2021-09-27
Differential fuzzing against rust version

catenacyber:
- Some changes to the test framework to be more consistent
- Adds a fuzz target for differential fuzzing
- Fixes an actual bug about "empty lines"

### DIFF
--- a/test/fuzz/fuzz_diff.c
+++ b/test/fuzz/fuzz_diff.c
@@ -17,12 +17,13 @@
 #include "htp/htp.h"
 #include "test/test.h"
 #include "fuzz_htp.h"
+#include "htp/htp_private.h"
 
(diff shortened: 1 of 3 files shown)


test/fuzz/fuzz_diff.c
```diff
@@ -0,0 +1,351 @@
...
+        }
+    }
+
+    htp_connp_close(connp, NULL);
+    return connp;
+}
+
+void* libhtprsFuzzRun(const uint8_t *Data, uint32_t Size);
+void* libhtprsFuzzConnp(void *);
+size_t htp_connp__rstx_size(void *);
```
catenacyber, line 257: Wonder if we should do nicer...

PR #332, opened 2021-10-10
host: ipv6 address is a valid host

catenacyber:
Replaces !329 with more portable includes

How can I test this on Windows/FreeBSD ?

PR #333, opened 2021-10-12
host: ipv6 address is a valid host

catenacyber:
Replaces #332 with better includes for `inet_pton`


catenacyber:
Replaced by https://github.com/OISF/libhtp/pull/334

PR #335, opened 2021-10-17
host: ipv6 address is a valid host

catenacyber:
Replaces https://github.com/OISF/libhtp/pull/334 with better includes for freebsd


victorjulien:
Merged in #336, thanks!

PR #338, opened 2021-10-29
util: better uint8_t typing

catenacyber:
cf https://redmine.openinfosecfoundation.org/issues/4516
cc @jlucovsky

### DIFF
--- a/htp/htp_util.c
+++ b/htp/htp_util.c
@@ -1147,10 +1147,10 @@ void htp_utf8_validate_path(htp_tx_t *tx, bstr *path) {
  * @param[in] data
  * @return decoded byte
  */
-static int decode_u_encoding_path(htp_cfg_t *cfg, htp_tx_t *tx, unsigned char *data) {
-    unsigned int c1 = x2c(data);
-    unsigned int c2 = x2c(data + 2);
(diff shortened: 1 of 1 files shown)


jlucovsky:
This looks good but it leaves a combination of `uint8_t` and `unsigned char` in use.  If the combination is intentional, then no change is needed. But, if `unsigned char` is used where a `uint8_t` is used, perhaps those could also be changed even if there are not any implicit size conversion issues?

catenacyber:
> But, if unsigned char is used where a uint8_t is used, perhaps those could also be changed even if there are not any implicit size conversion issues?

I guess I prefer `uint8_t` but I do not know if there is a good reason... Maybe because the size is explicit...

victorjulien:
> > But, if unsigned char is used where a uint8_t is used, perhaps those could also be changed even if there are not any implicit size conversion issues?
> 
> I guess I prefer `uint8_t` but I do not know if there is a good reason... Maybe because the size is explicit...

I definitely prefer explicit types over the older ones. So I'm all for replacing `unsigned char` with `uint8_t` and making similar conversions for other types.

catenacyber:
Replaced by https://github.com/OISF/libhtp/pull/339

PR #339, opened 2021-11-23
util: better uint8_t typing

catenacyber:
cf https://redmine.openinfosecfoundation.org/issues/4516

Replaces #338 with using `uint8_t` everywhere instead of `unsigned char` once


victorjulien:
Merged in #345, thanks!

PR #340, opened 2021-12-07
header: do not stop on null byte

catenacyber:
Do not stop on request headers, as is already done
on response headers, and in the libhtp rust version

The differential fuzzing found the difference on test case 68-invalid-request-header.t


victorjulien:
Merged in #345, thanks!

PR #349, opened 2022-02-14
headers: trim white spaces for responses values

catenacyber:
ping @victorjulien 

This is to test if anything breaks.

With this, differential fuzzing with the rust version should find no differences in the corpus with the tests files

### DIFF
--- a/htp/htp_response_generic.c
+++ b/htp/htp_response_generic.c
@@ -221,6 +221,12 @@ htp_status_t htp_parse_response_header_generic(htp_connp_t *connp, htp_header_t 
             break;
         }
     }
+    // Ignore LWS after field-content.
+    prev = value_end - 1;
+    while ((prev > value_start) && (htp_is_lws(data[prev]))) {
+        prev--;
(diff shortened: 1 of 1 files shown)


catenacyber:
Looks like @cccs-rtmorti did the opposite change in rust :
https://github.com/cccs-rtmorti/libhtp-rs/commit/755a0e5d46bf8fc4209a18bff96d46f1f0020e65

cccs-rtmorti:
> Looks like @cccs-rtmorti did the opposite change in rust : [cccs-rtmorti/libhtp-rs@755a0e5](https://github.com/cccs-rtmorti/libhtp-rs/commit/755a0e5d46bf8fc4209a18bff96d46f1f0020e65)

Yes, but this was only an attempt to match the existing libhtp behaviour. That change is easily reverted, and the rust version can go back to always trimming leading / trailing whitespace in header names and values.

catenacyber:
@cccs-rtmorti could you go back to the always trimming behavior ? Thanks

PR #356, opened 2022-05-23
htptest: Fixed an issue that could not be compiled.

pixclk:


catenacyber:
Could you please explain your problem with more details ?

pixclk:
Hello, thanks for your reply. I wanted to try this test before, but I found some problems that would cause compilation failed, and I tried to fix them:
1: struct submember reference error at line 133 and 143.
2: htp_server_personality_t enumeration error at line 557.
3: callback function (htp_config_register_response_complete) parameter error at line 559.
4: The internal private structure htp_connp_t is referenced but the corresponding header is not installed in the system directory, my solution is use the project header directly.

catenacyber:
How are you getting compilation failed ? (how can I reproduce ?)
What is your environment ?

PR #357, opened 2022-05-31
decompression: use less function pointers

catenacyber:
As it is always the same function anyways

### DIFF
--- a/htp/htp_decompressors.h
+++ b/htp/htp_decompressors.h
@@ -55,9 +55,7 @@ typedef struct htp_decompressor_t htp_decompressor_t;
 #define DEFLATE_MAGIC_2         0x8b
 
 struct htp_decompressor_t {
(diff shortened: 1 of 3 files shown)


htp/htp_decompressors.h
```diff
@@ -55,9 +55,7 @@ typedef struct htp_decompressor_t htp_decompressor_t;
 #define DEFLATE_MAGIC_2         0x8b
 
 struct htp_decompressor_t {
-    htp_status_t (*decompress)(htp_decompressor_t *, htp_tx_data_t *);
```
victorjulien, line 58: this is a public or private struct? If it is public we can't remove from it.

catenacyber: Indeed, restoring them

catenacyber:
Replaced by https://github.com/OISF/libhtp/pull/359

PR #358, opened 2022-07-07
htp: trim white space of invalid foloding for first header

catenacyber:
as does the rust version


victorjulien:
Merged in #365, thanks!

PR #359, opened 2022-07-31
Compression lessfnptrs v2

catenacyber:
decompression: use less function pointers as it is always the same function anyways

Replaces #357 with keeping the exported struct fields even if we no longer use them


catenacyber:
Replaced by https://github.com/OISF/libhtp/pull/360

Issue #361, opened 2022-08-02
Response body data buffering and callback bug


cccs-sadugas:
We found a bug where the response body data callback is getting called repeatedly with the same (ever growing) data.

Internally, I think this occurs when `RES_LINE` state treats incoming data as a response body because it can't find a valid response line.

**Steps to reproduce**
```
<<<
1
2
<<<
3
4
```

**Debug output**
```
htp_connp_res_data(connp->out_status 1)

htp_connp_res_data: ptr 0xFFFFFFFF2664 offset 0 len 3
0lx  31 0a 32                                          |1.2|

htp_connp_res_data: out state=RES_IDLE, progress=NULL
[LOG] Unable to match response to request
picked up response w/o requesthtp_connp_res_data: out state=RES_LINE, progress=RES_LINE

htp_connp_RES_LINE: ptr 0xFFFFFFFF2664 offset 0 len 2
0lx  31 0a                                             |1.|


htp_tx_res_process_body_data_ex: ptr 0xFFFFFFFF2664 offset 0 len 2
0lx  31 0a                                             |1.|

htp_connp_res_data: out state=RES_LINE, progress=RES_LINE
htp_connp_res_data: returning HTP_STREAM_DATA
htp_connp_res_data(connp->out_status 9)

htp_connp_res_data: ptr 0xFFFFFFFF266c offset 0 len 3
0lx  33 0a 34                                          |3.4|

htp_connp_res_data: out state=RES_LINE, progress=RES_LINE

htp_connp_RES_LINE: ptr 0xFFFFFFFF2f80 offset 0 len 3
0lx  32 33 0a                                          |23.|


htp_tx_res_process_body_data_ex: ptr 0xFFFFFFFF2f80 offset 0 len 3
0lx  32 33 0a                                          |23.|

htp_connp_res_data: out state=RES_LINE, progress=RES_LINE
htp_connp_res_data: returning HTP_STREAM_DATA
htp_connp_req_data(connp->in_status 2)

htp_connp_req_data: ptr NULL len 0
htp_connp_req_data: in state=REQ_FINALIZE, progress=NOT_STARTED

PROBING request finalize: ptr NULL len 0
htp_connp_req_data: in state=REQ_IDLE, progress=NULL
htp_connp_req_data: returning HTP_STREAM_DATA
htp_connp_res_data(connp->out_status 2)

htp_connp_res_data: ptr NULL len 0
htp_connp_res_data: out state=RES_LINE, progress=RES_LINE

[...]

PR #362, opened 2022-08-02
response: fix RES_LINE body data handling

cccs-sadugas:
This PR addresses the bug report #361 and also adds one test case that would fail on `0.5.x`.

The `out_buf` that is consolidated and used to read response data as provided by the user should be cleared to avoid growing indefinitely. This happens when the `RES_LINE` state cannot find a valid response line and treats user provided response data as a response body.

### Testing/QA

I ran `make check` to make sure existing tests still pass.


test/test_main.cpp
```diff
@@ -2088,3 +2089,60 @@ TEST_F(ConnectionParsing, Expect100) {
...
+    ASSERT_EQ(HTP_RESPONSE_COMPLETE, tx->response_progress);
+
+    struct ResponseBodyDataCallback *user_data = (struct ResponseBodyDataCallback *) htp_tx_get_user_data(tx);
+    ASSERT_TRUE(user_data);
+
+    char *response_body = strndup((const char *) user_data->data, user_data->len);
+    ASSERT_TRUE(response_body);
+
+    EXPECT_EQ(6, user_data->len);
+    EXPECT_STREQ("1\n23\n4", response_body);
```
catenacyber, line 2143: Should we somehow test that we first get "1 2" then "3 4" ?

htp/htp_response.c
```diff
@@ -1073,6 +1073,7 @@ htp_status_t htp_connp_RES_LINE(htp_connp_t *connp) {
 
                 connp->out_current_consume_offset = connp->out_current_read_offset;
                 htp_status_t rc = htp_tx_res_process_body_data_ex(connp->out_tx, data, len + chomp_result);
+                htp_connp_res_clear_buffer(connp);
```
catenacyber, line 1076: Nice, did you look for other missing calls to `htp_connp_res_clear_buffer` ?

cccs-sadugas: The `_FINALIZE()` state, which also calls the response body data hook clears this buffer. I haven't looked everywhere though. This should fix the bug we are experiencing.

test/test_main.cpp
catenacyber: Should we wait for a better test ?
 
(The fix looks good to me)

cccs-sadugas: You can go ahead and merge this PR while I work on a better test.

PR #360, opened 2022-07-31
Compression lessfnptrs v3

catenacyber:
decompression: use less function pointers as it is always the same function anyways

Replaces #357 with keeping the exported struct fields even if we no longer use them
Replaces #359 with changing unit test use of API

PR #364, opened 2022-08-31
Compression lessfnptrs v4

catenacyber:
decompression: use less function pointers as it is always the same function anyways

Replaces #360 fixing compilation warning about cast...


victorjulien:
Merged in #365, thanks!

PR #366, opened 2022-09-13
decompression: fix coverity warning

catenacyber:
about uninitialized field d.is_last


catenacyber:
Strange that the warning is new, as the bug is not new...

PR #367, opened 2022-10-28
Update htp_parsers.c

lrn0409:
connp->in_tx->request_auth_username need to set null,Otherwise, it may cause crash


catenacyber:
Thanks for this PR

> Otherwise, it may cause crash

Could you explain how it would crash ?
And provide a test about this crash being prevented ?

PR #370, opened 2022-11-11
fuzz: fix a null dereference in a diff report

catenacyber:
cc @cccs-rtmorti


victorjulien:
Merged in #373, thanks!

PR #371, opened 2022-11-14
htp: fixes warning about bad delimiter in URI

catenacyber:
With allow_space_uri, the protocol was checked for bad delimiter instead of the URI

See https://github.com/OISF/suricata/pull/8134

### DIFF
--- a/htp/htp_request_generic.c
+++ b/htp/htp_request_generic.c
@@ -379,7 +379,17 @@ htp_status_t htp_parse_request_line_generic_ex(htp_connp_t *connp, int nul_termi
             // special case: even though RFC's allow only SP (0x20), many
             // implementations allow other delimiters, like tab or other
             // characters that isspace() accepts.
-            while ((pos < len) && (!htp_is_space(data[pos]))) pos++;
+            pos = len - 1;
+            while ((pos > start) && (!htp_is_space(data[pos]))) pos--;
+        } else {
+            // reset bad_delim found in protocol part
+            bad_delim = 0;
+            for (size_t i = start; i < pos; i++) {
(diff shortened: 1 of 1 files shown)


htp/htp_request_generic.c
```diff
@@ -379,7 +379,16 @@ htp_status_t htp_parse_request_line_generic_ex(htp_connp_t *connp, int nul_termi
             // special case: even though RFC's allow only SP (0x20), many
             // implementations allow other delimiters, like tab or other
             // characters that isspace() accepts.
-            while ((pos < len) && (!htp_is_space(data[pos]))) pos++;
+            while ((pos > start) && (!htp_is_space(data[pos]))) pos--;
+        } else {
+            // reset bad_delim found in protocol part
+            bad_delim = 0;
+            for (size_t i = start; i < pos; i++) {
+                if (data[i] != 0x20 && htp_is_space(data[i])) {
```
victorjulien, line 388: I assume `0x20` is covered by `htp_is_space`. Are you adding it explicitly to avoid the func call?

victorjulien: hmm no I guess I'm failing to understand the logic

victorjulien: So the goal is to set a warning if there are "spaces" in the URI other than 0x20, right?

catenacyber: > So the goal is to set a warning if there are "spaces" in the URI other than 0x20, right?

It looks so

> I assume 0x20 is covered by htp_is_space.

Indeed, we want `htp_is_space` except for 0x20

victorjulien:
Merged in #373, thanks!

PR #375, opened 2022-12-30
fuzz_diff: Free the rust test object.

cccs-rtmorti:
When running the differential fuzzing I got segfaults because `libhtprsFreeFuzzRun()` expects a test structure, but is being passed a connp structure.


catenacyber:
I totally approve.
@victorjulien could we merge this ?

PR #378, opened 2023-03-16
readme: remove note on libhtp not being stable - v1

jufajardini:
It was noted in Suricata forum that it was still mentioned that libhtp wasn't stable. It is considered stable, at this point.

Seen on: https://forum.suricata.io/t/libhtp-unstable/3302


catenacyber (approved):
Fine for me

PR #383, opened 2023-03-29
htp: do not log content-encoding: none

catenacyber:
Ticket: #5757

https://redmine.openinfosecfoundation.org/issues/5757

### DIFF
--- a/htp/htp_transaction.c
+++ b/htp/htp_transaction.c
@@ -1453,7 +1453,7 @@ htp_status_t htp_tx_state_response_headers(htp_tx_t *tx) {
                                 "Compression bomb: multiple encoding with lzma");
                         break;
                     }
-                } else if (bstr_util_cmp_mem(tok, tok_len, "inflate", 7) == 0) {
+                } else if (bstr_util_cmp_mem(tok, tok_len, "inflate", 7) == 0 || bstr_util_cmp_mem(tok, tok_len, "none", 4) == 0) {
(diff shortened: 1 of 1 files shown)


htp/htp_transaction.c
```diff
@@ -1453,7 +1453,7 @@ htp_status_t htp_tx_state_response_headers(htp_tx_t *tx) {
                                 "Compression bomb: multiple encoding with lzma");
                         break;
                     }
-                } else if (bstr_util_cmp_mem(tok, tok_len, "inflate", 7) == 0) {
+                } else if (bstr_util_cmp_mem(tok, tok_len, "inflate", 7) == 0 || bstr_util_cmp_mem(tok, tok_len, "none", 4) == 0) {
```
victorjulien, line 1456: should these comparisons be stricter? I seems like "infateme" and "noneother" would match here as well due to the limited length check

victorjulien:
Merged in #384, thanks!

htp/htp_transaction.c
catenacyber: Do you want to ?

victorjulien: Wonder if it can lead to impedance mismatch between client and suricata? Suri thinks lzma, client something else?

catenacyber: I bet something can be crafted, but not easily

PR #385, opened 2023-04-07
response: headers wait to know if folding comes

catenacyber:
Previously, we fell back on HTTP/0.9 if there was a missing protocol except if the following line cintained a colon.

This makes libhtp stricter to consider a transaction as 0.9 by only accepting if we have spaces after the request line

Meant to fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=54982&q=label%3AProj-libhtp

### DIFF
--- a/htp/htp_response.c
+++ b/htp/htp_response.c
@@ -926,7 +926,8 @@ htp_status_t htp_connp_RES_HEADERS(htp_connp_t *connp) {
 
                 OUT_PEEK_NEXT(connp);
 
-                if (htp_is_folding_char(connp->out_next_byte) == 0) {
+                if (connp->out_next_byte != -1 &&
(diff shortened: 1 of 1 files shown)


catenacyber:
https://redmine.openinfosecfoundation.org/issues/5971

catenacyber:
Second commit fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=55007&q=label%3AProj-libhtp

victorjulien:
With this PR Suricata fails `http-close-header` in SV.

catenacyber:
Indeed missing also commit 3893c6a1 for responses

catenacyber:
Replaced by https://github.com/OISF/libhtp/pull/386

PR #387, opened 2023-04-26
response: skips lines before response line

catenacyber:
If we peek that one is coming

cc @jlucovsky


catenacyber:
for https://github.com/OISF/suricata/pull/8612

victorjulien:
The commit message isn't very clear, can you expand on what it does and why?

catenacyber:
> The commit message isn't very clear, can you expand on what it does and why?

Indeed, better commit message in https://github.com/OISF/libhtp/pull/389

PR #388, opened 2023-04-26
headers: log a warning for chunks extension

catenacyber:
cf https://github.com/OISF/libhtp/issues/83


catenacyber:
If this looks ok, I can do the suricata PR to get an event

PR #389, opened 2023-04-29
response: skips lines before response line

catenacyber:
Improve the logic for junk lines before response line, instead of treating them as body, try to peek if the next line begins with an H and may actually be the response line we were waiting for

If we do not see an H coming, we treat as if it were a HTTP/0.9 response to a HTTP1 request, that is a response without line and headers but just directly body.

for https://github.com/OISF/suricata/pull/8777

https://github.com/OISF/libhtp/pull/387 with commit message improved


victorjulien:
@jlucovsky

victorjulien:
Merged in #392, thanks!

PR #395, opened 2023-06-26
log: resist allocation failure

catenacyber:
If we cannot store the log message, simply return

As found by nallocfuzz


victorjulien (approved):

victorjulien:
Merged in #398, thanks!

PR #401, opened 2023-07-12
fix spelling issue

satta:
Automated quality checks in Debian showed a misspelled word in the log output.

### DIFF
--- a/htp/htp_util.c
+++ b/htp/htp_util.c
@@ -275,7 +275,7 @@ int64_t htp_parse_content_length(bstr *b, htp_connp_t *connp) {
     while ((pos < len) && (data[pos] < '0' || data[pos] > '9')) {
         if (!htp_is_lws(data[pos]) && connp != NULL && r == 0) {
             htp_log(connp, HTP_LOG_MARK, HTP_LOG_WARNING, 0,
-                    "C-L value with extra data in the beginnning");
+                    "C-L value with extra data in the beginning");
             r = -1;
         }
         pos++;


htp/htp_util.c
```diff
@@ -275,7 +275,7 @@ int64_t htp_parse_content_length(bstr *b, htp_connp_t *connp) {
     while ((pos < len) && (data[pos] < '0' || data[pos] > '9')) {
         if (!htp_is_lws(data[pos]) && connp != NULL && r == 0) {
             htp_log(connp, HTP_LOG_MARK, HTP_LOG_WARNING, 0,
-                    "C-L value with extra data in the beginnning");
+                    "C-L value with extra data in the beginning");
```
catenacyber, line 278: Does this require a Suricata change as well ?

satta: Not sure why it should...? I don't think so.

catenacyber: Suricata does app-layer events based on some logs...

It does not look there is this one.
Should there be ?

satta: Interesting! Well, I have not found the string `beginnning` anywhere in the Suricata source code, so I don't think there is an impact:
```
[satta@BLN04NB0421:~/tmp/suricata] $ git grep beginnn | wc -l
0
```

catenacyber: Indeed, there is none now.

But would you want an app-layer event when this log is triggered ?

satta: Good question. It is indeed a bit unusual to have leading non-numeric and non-whitespace content in front of a numeric value in this header, which according to the RFCs, should only contain digits -- even if the trailing content is skipped in the parsing.
OTOH no one has asked for such an event so far, and I would be reluctant to just introduce something for no real reason (YAGNI). I think we should merge this spelling thing and probably file a ticket for a potential event.

catenacyber: Ok for merging this

But how did you find this spelling mistake if it was not investigating some case where the app-layer event would have been useful ?

satta: Debian has a static checker, [Lintian](https://lintian.debian.org/), that checks [a zillion things](https://lintian.debian.org/tags) in source and binary packages. I was preparing a new [Debian upload for libhtp 0.5.45](https://tracker.debian.org/news/1443674/accepted-libhtp-10545-1-source-into-unstable/) and got the following:
```
$ lintian ../libhtp_0.5.45-1_amd64.changes
...
I: libhtp2: spelling-error-in-binary beginnning beginning [usr/lib/x86_64-linux-gnu/libhtp.so.2.0.0]
N: 
N:   Lintian found a spelling error in the given binary. Lintian has a list of common misspellings that it looks for. It does not have
N:   a dictionary like a spelling checker does.
N:   
N:   If the string containing the spelling error is translated with the help of gettext or a similar tool, please fix the error in the
N:   translations as well as the English text to avoid making the translations fuzzy. With gettext, for example, this means you should
N:   also fix the spelling mistake in the corresponding msgids in the *.po files.
N:   
N:   You can often find the word in the source code by running:
N:   
N:    grep -rw <word> <source-tree>
N:   
N:   This tag may produce false positives for words that contain non-ASCII characters due to limitations in strings.
N: 
N:   Visibility: info
N:   Show-Always: no
N:   Check: binaries/spelling
...
```
Usually such things get patched in the package and the patch is then forwarded upstream -- just as in this case. It looks like this misspelling was added in a recent version of the lintian tool and reported just now.

catenacyber (approved):
Thanks @satta

victorjulien:
Merged in #401, thanks!

PR #404, opened 2023-10-18

catenacyber:
Previously, we fell back on HTTP/0.9 if there was a missing protocol except if the following line contained a colon.

We make libhtp simpler by only switching to HTTP/0.9 if we have only spaces afterwards


catenacyber:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=54982&q=label%3AProj-libhtp&can=2

catenacyber:
Replaced by #405

PR #405, opened 2023-10-19

catenacyber:
Previously, we fell back on HTTP/0.9 if there was a missing protocol except if the following line contained a colon.

We make libhtp simpler by only switching to HTTP/0.9 if we have only spaces afterwards

#404 with passing tests

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=54982&q=label%3AProj-libhtp&can=2

### DIFF
--- a/htp/htp_request.c
+++ b/htp/htp_request.c
@@ -744,21 +744,15 @@ htp_status_t htp_connp_REQ_PROTOCOL(htp_connp_t *connp) {
     } else {
         // Let's check if the protocol was simply missing
         int64_t pos = connp->in_current_read_offset;
-        int afterspaces = 0;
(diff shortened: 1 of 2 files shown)


victorjulien:
What is the CIFuzz failure?

catenacyber:
> What is the CIFuzz failure?

This was a differential fuzzing because there was one PR for each repository (libhtp-rs)
Do you want me to repush a new PR so that CIFuzz runs new now that libhtp-rs is fixed ?

victorjulien:
Yeah that sounds good, thanks.

catenacyber:
Replaced by https://github.com/OISF/libhtp/pull/410

PR #412, opened 2023-12-21

catenacyber:
See https://redmine.openinfosecfoundation.org/issues/6643

https://github.com/OISF/libhtp/pull/410 with commit reworded to reference ticket 
S-V test is https://github.com/OISF/suricata-verify/pull/1561


catenacyber:
CIFuzz finds an input that has not the same behavior between libhtp-rs and this PR (but same with libhtp 0.5.x)

I think it is likely another variant of one of the known cases.
To be discussed with libhtp-rs

victorjulien (approved):

victorjulien:
Merged in #412, thanks!

PR #416, opened 2024-02-08
Http1 removetx 5921 v9

catenacyber:
https://redmine.openinfosecfoundation.org/issues/5921

Optimizations after 2fa869189278b0f380dfcdc23e0f205f91e4b354

PR #417, opened 2024-03-11
Http1 removetx 5921 v10

catenacyber:
#416 with dummy debug commit

### DIFF
--- a/htp/htp_connection.c
+++ b/htp/htp_connection.c
@@ -148,10 +148,11 @@ htp_status_t htp_conn_open(htp_conn_t *conn, const char *client_addr, int client
 htp_status_t htp_conn_remove_tx(htp_conn_t *conn, const htp_tx_t *tx) {
     if ((tx == NULL) || (conn == NULL)) return HTP_ERROR;
     if (conn->transactions == NULL) return HTP_ERROR;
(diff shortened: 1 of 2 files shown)


victorjulien:
Is the failing test above already enough proof?

catenacyber:
> Is the failing test above already enough proof?

Not yet, pushed another version passing tests

victorjulien:
It aborts, but again only in the unix socket test.

catenacyber:
Do you know which one of the 2 aborts ?

catenacyber:
Does the unix socket test free transactions differently ?

Anyways, I think we should not get this optimization soon...

victorjulien:
> Does the unix socket test free transactions differently ?

It shouldn't matter.
 
> Anyways, I think we should not get this optimization soon...

Agree. I will see if I can reproduce it manually after the releases next week.

catenacyber:
> > Does the unix socket test free transactions differently ?
> 
> It shouldn't matter.
> 

Sure, just trying to understand the issue, to reproduce and fix it.
It seems to happen because transaction are not freed in order...

catenacyber:
Replaced by https://github.com/OISF/libhtp/pull/418

PR #418, opened 2024-03-14
Http1 removetx 5921 v11

catenacyber:
#416 with resistance to having the last (but not first) transaction being freed : popping it, instead of replacing it by NULL, so that we can still use its index


victorjulien:
I get a Suricata SV failure with this:
```
===> http-sticky-server: Sub test #1: FAIL : expected 28 matches; got 26 for filter {'count': 28, 'match': {'event_type': 'alert', 'alert.signature_id': 2}}
```

victorjulien:
see https://github.com/OISF/suricata/actions/runs/8802614064

catenacyber:
Continued in https://github.com/OISF/libhtp/pull/423

PR #430, opened 2024-08-12
htp/table: only fetch element when needed

inashivb:
It is unnecessary to always fetch the element as it is only needed when there is a match. Get the list item only when the key candidate matches the key parameter and the element needs to be returned. This will ensure lesser calls to htp_list_get fn.


suricata-qa:
Information: QA ran without warnings. 

Pipeline 22052

catenacyber:
Is there a ticket for this ? A benchmark/profile ?
Should CentOS 7 be replaced with something newer ?

inashivb:
> Is there a ticket for this ?

Nope. Observed this while looking at your report for 7191.

> A benchmark/profile ?

None. Just seemed to make logical sense to me. wdyt?

> Should CentOS 7 be replaced with something newer ?

Sure. Almalinux? I can only find that we removed CentOS 7 in Suricata, I guess newer versions were already there in the CI there.

catenacyber:
> > A benchmark/profile ?
> 
> None. Just seemed to make logical sense to me. wdyt?

It makes sense.
But I wonder how the effect can be quantified...

> > Should CentOS 7 be replaced with something newer ?
> 
> Sure. Almalinux? I can only find that we removed CentOS 7 in Suricata, I guess newer versions were already there in the CI there.

@jasonish thoughts ?

suricata-qa:
Information: QA ran without warnings. 

Pipeline 22056

jasonish:
> @jasonish thoughts ?

I think we should have AlmaLinux 9 at least.  Maybe even 8 as well.

inashivb:
> I think we should have AlmaLinux 9 at least. Maybe even 8 as well.

ok. Thank you! Coming up in next rev

PR #439, opened 2025-03-07
response: end decompressors in chunked content

catenacyber:
When a response ends in the middle of a chunk, signal it to the decompressors to let them finalize the data they got so far, as is done with content-length.

Behaves the same as https://github.com/OISF/suricata/pull/12740

But this will require a QA rebaseline


suricata-qa:
ERROR:
 

ERROR: QA failed on SURI_TLPW1_files_sha256.
  

Pipeline 25028

victorjulien:
Merged in #440, thanks!

PR #449, opened 2025-03-19
request: do not fully error on missing request

catenacyber:
As it can happen in midstream case or gap
And the response may be handled as HTTP/0.9

Edit : 
QA has same errors as https://github.com/OISF/suricata/pull/12828 and next versions

### DIFF
--- a/htp/htp_request.c
+++ b/htp/htp_request.c
@@ -1006,7 +1006,8 @@ int htp_connp_req_data(htp_connp_t *connp, const htp_time_t *timestamp, const vo
     }
 
     // Sanity check: we must have a transaction pointer if the state is not IDLE (no inbound transaction)
-    if ((connp->in_tx == NULL)&&(connp->in_state != htp_connp_REQ_IDLE)) {
+    if ((connp->in_tx == NULL)&&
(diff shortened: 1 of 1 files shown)


suricata-qa:
WARNING:
 
 | field | baseline | test | % |
|--|--|--|--| 
| | SURI_TLPR1_stats_chk | | |
| .app_layer.error.http.parser | 536 | 485 | 90.49% | 

Pipeline 25257

catenacyber:
So, I think this is good and will require a QA rebaseline

victorjulien:
Can you create SV tests for these cases?

catenacyber:
> Can you create SV tests for these cases?

Which cases ? The remaining errors ? (not direct as it comes from TLPR)

victorjulien:
"As it can happen in midstream case or gap
And the response may be handled as HTTP/0.9"
This case ^

catenacyber:
> "As it can happen in midstream case or gap And the response may be handled as HTTP/0.9" This case ^

Spent too much time building https://github.com/OISF/suricata-verify/pull/2372
Not sure I found the original case that led to this commit message...

catenacyber:
New commit message in https://github.com/OISF/libhtp/pull/450

PR #464, opened 2026-05-06
fuzz: add additional harnesses to be consumed by OSS-Fuzz

DavidKorczynski:
The goal is to increase code coverage relative to the latest code coverage report
https://storage.googleapis.com/oss-fuzz-coverage/libhtp/reports/20260504/linux/src/report.html

Can confirm these new harnesses increase covered code, based on local runs.

Some adjustments are needed in OSS-Fuzz once these are merged.

### DIFF
--- a/test/fuzz/fuzz_htp_headers.c
+++ b/test/fuzz/fuzz_htp_headers.c
@@ -1,0 +1,88 @@
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "htp/htp.h"
+#include "htp/htp_private.h"
+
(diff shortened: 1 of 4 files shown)


catenacyber:
Thanks @DavidKorczynski 

I am not sure we are so interested in this : Suricata now only uses this libhtp C (it now uses an embedded libhtp in rust) for its main7 branch which is becoming EOL in the summer...
And Suricata does not use the multipart code that is in libhtp

That is why we do not bother for https://github.com/OISF/libhtp/issues/402

DavidKorczynski:
Sounds good, thanks for the quick reply @catenacyber

